Issue #13, opened 2017-11-10
Improve test coverage


ddysher:
From @walktall, current (11/10/2017) [test coverage](https://github.com/caicloud/nirvana/pull/10#issuecomment-343387587) is 75%. There is still room for improvement.

PR #20, opened 2017-11-11
feat(router): add tests for router

walktall:
**What this PR does**

- Add more tests for router.

- Refine some code when adding tests.

**why we need it**

Fix https://github.com/caicloud/nirvana/issues/13

@kdada @ddysher

### DIFF
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -2,198 +2,283 @@ package router
 
 import (
 	"context"
-	"encoding/json"
+	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 )
 
-func TestSplit(t *testing.T) {
-	paths := []string{
-		"/segments/segment/resources/resource",
-		"/segments/{segment}/resources/{resource}",
-		"/segments/{segment:[a-z]{1,2}}.log{temp}sss/paths/{path:*}",
-	}
-	results := [][]string{
-		{"/segments/segment/resources/resource"},
-		{"/segments/", "{segment}", "/resources/", "{resource}"},
-		{"/segments/", "{segment:[a-z]{1,2}}", ".log", "{temp}", "sss/paths/", "{path:*}"},
-	}
-	for i, p := range paths {
-		result, err := Split(p)
-		if err != nil {
-			t.Fatalf("Untracked error: %s", err.Error())
+var (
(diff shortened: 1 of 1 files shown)


walktall:
@kdada
 
`Parse("localhost")`
`Parse("api/v1")`

Should return an error.

router/router.go
```diff
@@ -289,9 +289,9 @@ type ExpSegment struct {
 	Key string
 }
 
-// ParseExpSegment parses a regexp segment to ExpSegment.
-func ParseExpSegment(exp string) (*ExpSegment, error) {
-	if !strings.HasPrefix(exp, "{") || !strings.HasSuffix(exp, "}") {
+// parseExpSegment parses a regexp segment to ExpSegment.
+func parseExpSegment(exp string) (*ExpSegment, error) {
```
ddysher: Please use a different PR for this

router/router.go
```diff
@@ -289,9 +289,9 @@ type ExpSegment struct {
...
+	if !strings.HasSuffix(exp, "}") || !strings.HasPrefix(exp, "{") {
```
ddysher: why changing the order?

walktall: The first `{` checking was done before the execution going into this func. So it was redundant.
In fact, we only need to check `}`.

router/router.go
walktall: Then it is better to leave it to @kdada I think.

walktall:
@ddysher Got 1 bug. `localhost` and `api/v1` will pass the examination. Already @kdada in comment.

kdada:
@walktall 

Why do you consider it's a bug?

walktall:
@kdada 
`Parse` didn't report error, but `Match` failed.

Jimexist:
is this still WIP?

walktall:
@Jimexist Yes.

ddysher:
Need to update after conclusion from https://github.com/caicloud/nirvana/issues/24.

walktall:
@ddysher @Jimexist PTAL

Jimexist:
@walktall tests failed

walktall:
ping @Jimexist

Jimexist:
merge to move forward.

PR #22, opened 2017-11-14
fix(router): ignore empty string segment in splitting

kdada:
**What this PR does**
`Split` creates a useless empty string segment when there is nothing before char '{'. It's a bug.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -194,7 +194,9 @@ func Split(path string) ([]string, error) {
 		case '{':
 			braceCounter++
 			if braceCounter == 1 {
-				result = append(result, path[lastElementPos:i])
+				if i > lastElementPos {
+					result = append(result, path[lastElementPos:i])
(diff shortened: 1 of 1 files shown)


Jimexist:
when you changed an implementation without changing unit tests and the unit tests passed - it means your unit tests were not correctly coverring.

ddysher:
Agree with @Jimexist; let's checkin [unittest](https://github.com/caicloud/nirvana/pull/20) change first.  We can't merge this unless unittest is added to cover this bug.

walktall:
@ddysher @Jimexist The test for this bug will be added [here](https://github.com/caicloud/nirvana/pull/20).

ddysher:
Discussed offline, test is already covered [here](https://github.com/caicloud/nirvana/pull/20).  We'll merge this PR first.

PR #27, opened 2017-11-17
Web framework preview

kdada:
**What this PR does**
The pr contains main logic of web framework. It's not the final version. Don't merge it now.


web/executor.go
```diff
@@ -0,0 +1,94 @@
...
+	out      outGenerator
+}
+
+func newExecutor(d *definition.Definition) (router.Executor, error) {
+	method := definition.HTTPMethodFor(d.Method)
+	if method == "" {
+		return nil, fmt.Errorf("no HTTP method for %s", d.Method)
+	}
+	code := definition.HTTPCodeFor(d.Method)
+	if method == "" {
```
Jimexist: code?

kdada: Yes.

definition/definition.go
```diff
@@ -0,0 +1,132 @@
...
+	Number    Kind = "Number"
+	Numbers   Kind = "Numbers"
+	String    Kind = "String"
+	Strings   Kind = "Strings"
+	Binary    Kind = "Binary"
+	Boolean   Kind = "Boolean"
+	Booleans  Kind = "Booleans"
+	Date      Kind = "Date"
+	Dates     Kind = "Dates"
+	DateTime  Kind = "DateTime"
```
walktall: What is the difference between `DateTime` and `Date`?

kdada: I'm adding comments now.

examples/api-basic/api/v1/app.go
```diff
@@ -0,0 +1,43 @@
...
+var app = definition.Descriptor{
+	Path: "/applications",
+	Definitions: []definition.Definition{
+		{
+			Method:   definition.Create,
+			Function: application.CreateApplication,
+			Parameters: []definition.Parameter{
+				{
+					Source: definition.Body,
+					Type:   reflect.TypeOf(&application.ApplicationV1{}),
```
walktall: I'd rather write another `Function` to handle the convert process.

definition/definition.go
```diff
@@ -0,0 +1,132 @@
...
+	Description string
+	Instance    interface{}
+}
+
+type Parameter struct {
+	Source Source
+	Name   string
+	// Optional. If you want to override the type in function parameter,
+	// you can specify it here.
+	Type      reflect.Type
```
walktall: I think this `Type` will add extra burden for users...
Since it was used to handle API compatibility which is uncommon in normal cases.

definition/definition.go
```diff
@@ -0,0 +1,132 @@
...
+	Examples    []Example
+}
+
+type Descriptor struct {
+	Path        string
+	Consumes    []string
+	Produces    []string
+	Middlewares []router.Middleware
+	Definitions []Definition
+	Children    []Descriptor
```
walktall: `Children` is a strange name.
Prefer `SubGroup`?

examples/api-basic/api/v1/operators/converter.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"reflect"
+
+	"github.com/caicloud/nirvana/definition"
+	"github.com/caicloud/nirvana/examples/api-basic/application"
+)
+
+func operater(f interface{}) definition.Operator {
+	value := reflect.ValueOf(f)
```
walktall: It is not good for a framework to make user write `reflect` code...especially in their business code lines.

web/content.go
```diff
@@ -0,0 +1,370 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"reflect"
+	"strconv"
+)
+
+// acceptTypeAll indicates a accept type from http request.
+// It means client can receive any content.
+// Request content type in header "Content-Type" must not set to "*/*".
```
walktall: I think server should handle this case...

web/content.go
```diff
@@ -0,0 +1,370 @@
...
+type JSONSerializer struct{}
+
+// ContentType returns json MIME type.
+func (s *JSONSerializer) ContentType() string {
+	return "application/json"
+}
+
+// Consume unmarshals json from r into v.
+func (s *JSONSerializer) Consume(r io.Reader, v interface{}) error {
+	data, err := ioutil.ReadAll(r)
```
walktall: Why not use `json.Decoder`?

web/content.go
```diff
@@ -0,0 +1,370 @@
...
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(data, v)
+}
+
+// Produce marshals v to json and write to w.
+func (s *JSONSerializer) Produce(w io.Writer, v interface{}) error {
+	data, err := json.Marshal(v)
```
walktall: Why not just use `json.Encoder`?

web/content.go
kdada: Why?

examples/api-basic/api/v1/operators/converter.go
kdada: It's an example!

examples/api-basic/api/v1/app.go
kdada: You think 
```go
func AConverter(ctx context.Context, filed string, xxx sometype, obj *A) (*A,error)
func BConverter(ctx context.Context, filedxx int, obj *B) (*B,error)
```
is better than
```go
func AConverter(ctx context.Context, obj *A)  (*A,error)
func BConverter(ctx context.Context, obj *B)  (*B,error)
```
?

web/context.go
```diff
@@ -0,0 +1,276 @@
...
+
+// clear clears request and response.
+func (c *httpCtx) clear() {
+	c.Context = nil
+	c.container.request = nil
+	c.container.query = nil
+	c.response.writer = nil
+}
+
+// Value returns itself when key is contextKeyUnderlyingHttpContext.
```
walktall: It is not possible for a outside pkg passing `contextKeyUnderlyingHttpContext` to `Value`.

web/content.go
kdada: The world does not only have json. There are xml, files, and many strange types.

web/content.go
kdada: Yes, that's right.

web/context.go
kdada: External packages should not use httpCtx.
There are three utils for external packages:
```go
func HTTPRequest(ctx context.Context) *http.Request 
func HTTPResponseWriter(ctx context.Context) ResponseWriter 
func HTTPValueContainer(ctx context.Context) ValueContainer 
```

examples/api-basic/api/v1/app.go
walktall: Yes!!!!!

examples/api-basic/api/v1/operators/converter.go
walktall: Example should be easy to read for potential users...

web/context.go
walktall: Yes, but users can see it it in comments...

web/content.go
```diff
@@ -0,0 +1,337 @@
...
+type Prefab interface {
+	// Name returns prefab name.
+	Name() string
+	// Type is instance type.
+	Type() reflect.Type
+	// Make makes an instance.
+	Make(ctx context.Context) (interface{}, error)
+}
+
+var prefabs = map[string]Prefab{}
```
walktall: Should we protect it with a lock?

walktall:
Should we put web into root?

web/web.go
```diff
@@ -0,0 +1,231 @@
...
+	ServeHTTP(resp http.ResponseWriter, req *http.Request)
+}
+
+// NewDefaultServer creates a server with default config.
+// It uses three filters: RedirectTrailingSlash, ParseRequestForm, FillLeadingSlash.
+// It uses one modifier: FirstContextParameter.
+func NewDefaultServer() Server {
+	s := NewServer()
+	s.AddFilters(RedirectTrailingSlash(), ParseRequestForm(), FillLeadingSlash())
+	s.SetModifier(FirstContextParameter())
```
caitong93: If `FirstContextParameter()` is not set, server will panic. So `FirstContextParameter()` modifier should always be added.
The problem is `NewServer`  does not set this modifier. Please fix it.

kdada: I'll add more comments to function `NewServer()`. The server is not used for common scene. It provides a complete capacity to customize a specific server.

web/web_test.go
```diff
@@ -0,0 +1,185 @@
...
+	t.Log(resp.header)
+	t.Logf("%s", resp.buf.Bytes())
+	if resp.code != 201 {
+		t.Fatalf("Response code should be 201, but got: %d", resp.code)
+	}
+	if resp.header == nil || resp.header.Get("Content-Type") != "application/json" {
+		t.Fatalf("Content-Type should be application/json, but got: %s", resp.header.Get("Content-Type"))
+	}
+	result := string(resp.buf.Bytes())
+	target := `{"name":"asdasd","namespace":"system","target":"nothing","target2":1,"target1":false}` + "\n"
```
walktall: 这个jsonencode出来不一定是这个顺序吧？

ddysher: @kdada please take a look later, I'll merge to proceed.

Issue #39, opened 2017-11-25
GitHub page and example projects


Jimexist:
**What is the issue about**:
GitHub pages should be set up along with two example projects:
- hello world
- todo app

**What happened**:
Nothing happened, I just figure that a more out-reaching stance this repo shall take.

**What you expected to happen**:
More open source usage, more tractions, etc.

**Anything else we need to know**:

- [x] skeleton #40 
- [ ] add status badge and add *fork me at github button* 
- [ ] choose a better theme
- [x] create `CNAME` and apply domain

ddysher:
nice!  Apart from examples, we also need to add more documentation, usage, etc.

Do you need a domain :)  http://gonirvana.org/ ?

ddysher:
As discussed offline, apart from a 'large' getting started example, we need a set of small example projects to demonstrate features. Below is a non-exhaustive list:

- [ ] hello-world example
- [ ] stream API
- [ ] error message
- [ ] openapi generation
- [ ] cli configuration
- [ ] tracing

@kdada to add more

PR #40, opened 2017-11-25
feat(docs): add gitbook

Jimexist:
**What this PR does**
add skeleton for docs and website

**why we need it**
for PR of course :-)

pertaining to #39


docs/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -euf -o pipefail
+gitbook build
+rm -rf ../docs
```
ddysher: is it safe to always remove the directory?

website/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -euf -o pipefail
+gitbook build
+rm -rf ../docs
```
ddysher: same here?

docs/build.sh
Jimexist: it is... because `/docs` is auto-generated and there's git to track history.

website/build.sh
Jimexist: it is... because `/docs` is auto-generated and there's git to track history.

Jimexist:
@ddysher 
> Are we going to put all docs under website?

`.md`s under `webiste`, generated gitbook under `/docs` so that GitHub pages may pickup

docs/build.sh
ddysher: yeah, it's auto-generated here, but it is quite a established rules to put docs under `docs` directory.

Issue #3, opened 2017-09-22
[Umbrella] API validation


ddysher:
Umbrella issue for tracking API validation tasks, scopes:
- [x] Research on requirements, e.g. what kind of validation do we need
- [x] Research on existing solutions
- [x] Proposal or demonstration on how validation should work in nirvana
- [x] Integration with basic framework
- [x] Testing

@walktall

ddysher:
@walktall Action Items after 4th meeting:
- LGTM on basic framework implementation (11/24/2017)
- Draft proposal on validation implementation (11/24/2017)

walktall:
How to implement sample:

```
// Package validator usage:

// type Application struct {
// 	Name      string `json:"name" validate:"required,printascii"`
// 	Namespace string `json:"namespace"`
// }

// Definitions: []definition.Definition{
// 	{
// 		Method:   definition.Create,
// 		Function: Handle,
// 		Parameters: []definition.Parameter{
// 			{
// 				Source:    definition.Query,
// 				Name:      "target1",
// 				Operators: []definition.Operator{validator.Var("gt=0,lt=10")},
// 			},
// 			{
// 				Source: definition.Body,
// 				Name:   "app",
// 				// Should we call this automatically?
// 				Operators: []definition.Operator{validator.Struct()},
// 			},
// 		},
// 	},
// },

package validator

import (
	"context"

	"gopkg.in/go-playground/validator.v9"
)

var std = validator.New()

func Var(tag string) func(ctx context.Context, object interface{}) (interface{}, error) {
	return func(ctx context.Context, object interface{}) (interface{}, error) {
		// TODO: error format
		return object, std.VarCtx(ctx, object, tag)
	}
}

func VarWithValue(other interface{}, tag string) func(ctx context.Context, object interface{}) (interface{}, error) {
	return func(ctx context.Context, object interface{}) (interface{}, error) {
		err := std.VarWithValueCtx(ctx, object, other, tag)
		return object, err
	}
}

func Struct() func(ctx context.Context, object interface{}) (interface{}, error) {
	return func(ctx context.Context, object interface{}) (interface{}, error) {
		return object, std.StructCtx(ctx, object)
	}
}

```

ddysher:
Action Items after 5th meeting:
- Integration with basic framework, 12/01/2017

ddysher:
Done 🚀

PR #46, opened 2017-12-01
feat(validator): basic validator

walktall:
**What this PR does**

Add basic validator.

**why we need it**

#3 


validator/validator.go
```diff
@@ -0,0 +1,61 @@
+// Package validator usage:
```
ddysher: This makes the file too verbose, can we use `doc.go` instead for documentation purpose?

validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+// 		Parameters: []definition.Parameter{
+// 			{
+// 				Source:    definition.Query,
+// 				Name:      "target1",
+// 				Operators: []definition.Operator{validator.Var("gt=0,lt=10")},
+// 			},
+// 			{
+// 				Source: definition.Body,
+// 				Name:   "app",
+// 				// Should we call this automatically?
```
ddysher: Is this still a question?

validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+
+	"gopkg.in/go-playground/validator.v9"
+)
+
+var std = validator.New()
+
+// Var returns an operator to validate a single variable using tag style validation and allows passing of contextual
+// validation validation information via context.Context.
```
ddysher: duplidate

validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+func Var(tag string) func(ctx context.Context, object interface{}) (interface{}, error) {
+	return func(ctx context.Context, object interface{}) (interface{}, error) {
+		// TODO: error format
```
ddysher: Please file an issue

validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+
+package validator
+
+import (
+	"context"
+
+	"gopkg.in/go-playground/validator.v9"
+)
+
+var std = validator.New()
```
ddysher: Please add documentation to point to std validation and mention that only std is supported now.

This can be done in another PR.

For example (or documentation), we need one for single variable and one for struct.

validator/validator.go
walktall: After offline discuss, we will call it explicit here.

validator/validator.go
walktall: OK, will add it in another README PR.

ddysher:
LGTM  @walktall There is a conflict with @zoumo on vendor package.

walktall:
@ddysher Conflict resolved.

PR #53, opened 2017-12-04
feat(errors): add errors

walktall:
**What this PR does**

Add errors pkg.

**why we need it**

Currently, errors is not ready to use.
This PR will make a standard error which should be used in future.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -18,38 +18,34 @@ package errors
 
 import (
 	"fmt"
-	"net/http"
 )
 
-type Error interface {
-	Reason() Reason
-	Error() string
-}
-
 // Reason is an enumeration of possible failure causes. Each Reason
-// must map to a single HTTP status code, but multiple reasons may map to the
-// same HTTP status code.
+// must map to a format which is a string containing ${formatArgu1}.
+// Exp:
+// Reason "kind:NotFound" may map to Format "${kindName} was not found".
+// Reason "Status:Sleep" may map to Format "${Name} is sleeping now"
 type Reason string
 
+// message was the result returned to client.
+// Exp:
+// {
+//   "message": "name japari is too short",
+//   "reason": "monitoring:CreateDashboardNameTooShort",
+//   "data": {
+//     "name": "japari"
+//   }
+// }
 type message struct {
-	// Required, a machine-readable description of the error.
-	Reason Reason `json:"reason"`
-	// Required when template is used in message or i18nMessage.
-	Data []interface{} `json:"data,omitempty"`
-	// Required for 4xx but optional for 5xx. Message is a human-readable description
-	// of the error. Message can be golang template.
-	Message string `json:"message"`
+	Reason  Reason            `json:"reason,omitempty"`
+	Message string            `json:"message"`
+	Data    map[string]string `json:"data,omitempty"`
 }
 
-// err is an error intended to be used by all APIs to return error to clients.
 type err struct {
-	// Suggested HTTP return code for this error, 0 if not set. This field is optional.
-	// Caller can choose to use this code or choose to use another error code for client.
-	code int
-	// Format is used to generate message.
+	message
+	code   int
 	format string
-	// Useful message.
-	message message
 }
 
 func (e *err) Code() int {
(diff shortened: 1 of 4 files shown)


errors/errors.go
```diff
@@ -57,43 +53,68 @@ func (e *err) Code() int {
...
+	return err
 }
 
-func ResourceNotFound(resource string) Error {
-	return (&formatter{
-		code:   http.StatusNotFound,
-		reason: Reason(http.StatusText(http.StatusNotFound)),
-		format: "%s is not found",
-	}).Format(resource)
+func expand(format string, v ...interface{}) (msg string, data map[string]string) {
```
ddysher: comment please

errors/errors.go
```diff
@@ -18,36 +18,32 @@ package errors
...
 )
 
+// Reason is an enumeration of possible failure causes. Each Reason
+// must map to a format which is a string contains ${xxx}.
+type Reason string
+
+// Error is the basic form of an error returned by nirvana.
 type Error interface {
-	Reason() Reason
+	Message() interface{}
```
ddysher: Returning `interface{}` loses observability on what constitutes a Message. Since this is probably the most important interface in error package, I'd prefer a struct or at least another interface type.

walktall: Since web framework need a
type Error interface {
	Message() interface{}
	Code() int
}
here, ref: https://github.com/caicloud/nirvana/blob/master/web/handler.go#L31

So this `Message()` must return interface{}.

errors/errors.go
```diff
@@ -57,43 +53,68 @@ func (e *err) Code() int {
...
+// New creates an error which can be used to generate this kind of error.
+// exp:
+// {
+//   "message": "name japari is too short",
+//   "reason": "monitoring:CreateDashboardNameTooShort",
+//   "data": {
+//     "name": "japari"
+//   }
+// }
+func New(code int, reason Reason, format string, a ...interface{}) error {
```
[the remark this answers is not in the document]
walktall: err will not be used here, since a exported func should not return a hidden struct type I think.
Will change it to Error instead.

errors/errors.go
ddysher: Having two `Error` interfaces at two different places is awful, do we have any other solution?

@walktall @kdada

errors/errors.go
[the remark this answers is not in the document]
ddysher: ops, forgot to remove `*err`...  what I mean is s/error/Error

errors/errors.go
walktall: web pkg can import errors pkg.

errors/errors.go
```diff
@@ -18,36 +18,35 @@ package errors
...
 
+// Reason is an enumeration of possible failure causes. Each Reason
+// must map to a format which is a string containing ${formatArgu1}.
+// Exp:
+// Reason "kind:NotFound" may map to Format "${kindName} was not found".
+// Reason "Status:Sleep" may map to Format "${Name} is sleeping now"
+type Reason string
+
+// Error is the basic form of an error returned by nirvana.
 type Error interface {
```
zoumo: The Error is confused with the built-in Error interface.
The similar concept in k8s is called [APIStatus](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L45:1)

ddysher:
@walktall please update the PR as discussed offline.

ddysher:
We need to provide well-defined errors to regulate error usage, e.g. https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go. To allow custom error message, you can leave a formatter method as discussed.

Flexibility comes with a cost of breaking things. I want error types to be well established.

walktall:
@ddysher I don't think so, a Factory is enough to restrict the error format and control it.
And a formatter method is not enough since format string is paired with reason.
Code will be like this if we add formatter and reason setter:

```
func (f Factory) SetReason(r Reason) *Factory {
	f.reason = r
	return &f
}

func (f Factory) SetFormat(s string) *Factory {
	f.format = s
	return &f
}

var BadRequest = NewFactory(400, Reason("badRequest"), "user need to check this field when they are using provided errors.BadRequest")

// In user's pkg, they still need to write
var BadRequest = errors.BadRequest.SetReason("events:badRequest").SetFormat("${name} mismatch")
// which is much more confused than 
var BadRequest = errors.NewFactory(400, errors.Reason("events:badRequest"), "${name} mismatch")
```

The actually error should not be set by a web framework, it is business staff. https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go this is error just for k8s but not for other cases and we are making a web framework but not another k8s repo.


Usage for errors pkg should be like this:

```
// User writes an error pkg in their own repo which import nirvana/errors, and

var BadRequest = errors.NewFactory(400, errors.Reason("events:badRequest"), "${name} mismatch")

```

If we want to control the format globally, we can create a caicloud/errors only used in our cases.

kdada:
Usage in my mind:
```
var ApplicationIsNotFound = errors.EntityNotFound.Factory("app-admin:application-module:application-not-found","server can't find application {applicationName} in partition {partitionName}")

func GetApplication() (*Application, error) {
    return ApplicationIsNotFound.New("mongodb", "test")
}

func main() {
    _, err := GetApplication()
    if ApplicationIsNotFound.Equal(err) {
        // Do something.
    }
}
```

errors/errors.go
```diff
@@ -18,82 +18,136 @@ package errors
...
+	x, ok := e.(*err)
+	if !ok {
+		return false
+	}
+	return f.code == x.code && f.reason == x.message.Reason && f.format == x.format
+}
+
+// Type maps to http code.
+// And it can be used to make an error factory.
+type Type struct {
```
zoumo: May StatusCode is more appropriate。

errors/errors.go
```diff
@@ -18,82 +18,114 @@ package errors
...
+// }
 type message struct {
-	// Required, a machine-readable description of the error.
-	Reason Reason `json:"reason"`
-	// Required when template is used in message or i18nMessage.
-	Data []interface{} `json:"data,omitempty"`
-	// Required for 4xx but optional for 5xx. Message is a human-readable description
-	// of the error. Message can be golang template.
-	Message string `json:"message"`
+	Reason  Reason            `json:"reason,omitempty"`
```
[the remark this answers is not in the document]
zoumo: Reason here is counterintuitive. Reason and Status Code is related to each other. They should be together.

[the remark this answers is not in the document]
walktall: message is used to as response body, so reason must be here.

errors/errors.go
walktall: Different status code means different type of errors, so I used Type here.

PR #70, opened 2017-12-14
feat(validator): add validator readme

walktall:
**What this PR does**

Add validator pkg README.md

**why we need it**

There is no README.md for validator pkg now.

### DIFF
--- a/validator/README.md
+++ b/validator/README.md
@@ -1,0 +1,80 @@
+# Validator
+
+## About
+
+We use [go-playground/validator.v9](https://github.com/go-playground/validator/tree/v9)
+which provides flexible and rich methods to validate structs and values for validation.
+
+## Design
+
+Validator makes use of `Operators` in `definition.Parameter` struct.
+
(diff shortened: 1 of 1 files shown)


validator/README.md
```diff
@@ -0,0 +1,78 @@
...
+```
+type Booking struct {
+	UserName string    `json:"name" validate:"required,printascii"`
+	CheckIn  time.Time `json:"checkIn" validate:"required"`
+	CheckOut time.Time `json:"checkOut" validate:"required,gtfield=CheckIn"`
+}
+
+func validateBooking(ctx context.Context, obj interface{}) (interface{}, error) {
+	// Note we can still use validator pkg to check before our custom validation code.
+	obj, err := validator.Struct()(ctx, obj)
```
supereagle: What is `validator`? A package, or a variable?

ddysher: @walktall The README doesn't provide necessary background.  I'm gonna merge it first.  We'll do documentation during end of Dec.

PR #71, opened 2017-12-15
feat(example): add example for stream api

kdada:
**What this PR does**
Add example for stream api.

**why we need it**
To show how we transfer big data.

@Jimexist


examples/stream/stream/create.go
```diff
@@ -0,0 +1,37 @@
...
+
+package stream
+
+import (
+	"bytes"
+	"context"
+	"io"
+)
+
+func Stream(ctx context.Context) (io.Reader, error) {
```
ddysher: The framework is responsible to read and stream the data, and the 'interface' we provide to framework user is `definition.Data`?

kdada: Yes. Whatever the type of the returned value is, for users, it's `Data`.

examples/stream/api/v1/stream.go
```diff
@@ -0,0 +1,45 @@
...
+	register(app)
+}
+
+var app = definition.Descriptor{
+	Path: "/stream",
+	Definitions: []definition.Definition{
+		{
+			Method:   definition.Create,
+			Function: stream.Stream,
+			Produces: []string{"application/somedata"},
```
Jimexist: what does this mean?

kdada: I just define a content type to explain how to use producer to handle `io.Reader`.

Jimexist:
merge to move forward

PR #87, opened 2017-12-19
feat(lint): add unused, and then replace with megacheck

Jimexist:
**What this PR does**
add `unused` check and then replace with metacheck

**why we need it**

why those code were not used?

`megacheck` = `unused` + `staticcheck` + `gosimple`


ddysher:
@zoumo to take a look

cmd/flag-gen/types/types.go
```diff
@@ -16,27 +16,24 @@ limitations under the License.
 
 package types
 
-import (
-	"time"
-)
-
 // go:generate flag-gen -i github.com/caicloud/nirvana/cmd/flag-gen/types -o github.com/caicloud/nirvana/cli
 
+// TODO: why not used?
```
zoumo: flag-gen read this struct to generate code

cli/config_test.go
```diff
@@ -91,9 +91,10 @@ func TestIsSet(t *testing.T) {
 }
 
 func TestGet(t *testing.T) {
-	type args struct {
-		key string
-	}
+	// TODO: why not used?
```
zoumo: auto generated by vscode plugin, can be deleted

cmd/flag-gen/types/types.go
ddysher: then we have to trick the tool to not warning on this?

Jimexist: fixed by adding `// nolint:`

cli/config_test.go
Jimexist: removed

PR #93, opened 2017-12-21
Refactor directory structure

kdada:
**What this PR does**
- fix(*): fix headers for *.sh and *.go
- refactor(*): refactor directory structure
- feat(log): implement a basic std logger

The basic logger is very rough. It will be enhanced in the future.


utils/profiling/profiling.go
```diff
@@ -51,8 +50,8 @@ func pprofHandler(w http.ResponseWriter, r *http.Request) {
 	pprofIndex(w, r)
 }
 
-// Install adds the Profiling webservice to the given mux.
-func (d profiling) Install(s web.Server) {
+// Install adds the Profiling serviceservice to the given mux.
```
supereagle: `serviceservice`, typo?

log/std.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (l *stdLogger) exitIfFatal(severity Severity) {
+	if severity == SeverityFatal {
+		os.Exit(1)
+	}
+}
+
+func (l *stdLogger) write(data string) {
+	if data[len(data)-1] != '\n' {
```
Jimexist: what if i logged empty string?

kdada: It's impossible.  `data` has `header` at least.
```
INFO  1222-18:01:16.566+08 log_test.go:94 | 
```

kdada: It's impossible.  `data` has `header` at least.
```
INFO  1222-18:01:16.566+08 log_test.go:94 | 
```

PR #102, opened 2018-01-13
feat(service): let consumers and producers support string and []byte

kdada:
**What this PR does**
Now, Type `string` and `[]byte` can be used to receive data from http body.

**why we need it**
In some occasion, we just want get body string rather than a struct.

**Release note**
```release-note
Let consumers and producers support string and []byte.
```

### DIFF
--- a/service/content.go
+++ b/service/content.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"io"
+	"io/ioutil"
 	"reflect"
 	"strconv"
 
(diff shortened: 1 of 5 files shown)


service/content.go
```diff
@@ -129,16 +128,118 @@ func (s *NoneSerializer) ContentType() string {
...
+
+// ContentType returns form data MIME type.
+func (s *FormDataConsumer) ContentType() string {
+	return definition.MIMEFormData
+}
+
+// Consume reads data and converts it to string, []byte.
+func (s *FormDataConsumer) Consume(r io.Reader, v interface{}) error {
+	return s.consume(s.ContentType(), r, v)
 }
```
yejiayu: Just so easy for parsing?

kdada: It works only if you don't register a `service.ParseRequestForm()` filter. If the filter is installed, the body of a `multipart/formdata` request would be parsed and you will get nothing by the consumer.

In most occasion, you should not use the consumer to get request body. `definition.File` and `definition.Form` is recommended.

PR #108, opened 2018-01-17
feat(source): add default tag and set default value

yejiayu:
- Add `default` source tag. 
- All basic types of params have a default value.
Fixes #

/assign @kdada 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/service/executor.go
+++ b/service/executor.go
@@ -361,8 +361,12 @@ func (e *executor) Execute(ctx context.Context) (err error) {
 		if err != nil {
 			return writeError(ctx, e.producers, err)
 		}
-		if result == nil && p.defaultValue != nil {
-			result = p.defaultValue
+		if result == nil {
+			if p.defaultValue != nil {
+				result = p.defaultValue
+			} else {
+				result = reflect.Zero(p.targetType).Interface()
+			}
 		}
 		for _, operator := range p.operators {
 			result, err = operator.Operate(ctx, p.name, result)
--- a/service/source.go
+++ b/service/source.go
@@ -357,6 +357,21 @@ func (g *PrefabParameterGenerator) Generate(ctx context.Context, vc ValueContain
 // }
 type AutoParameterGenerator struct{}
 
+type autoTagParams map[paramsKey]string
+type paramsKey string
+
+const (
+	keyDefault paramsKey = "default"
+)
+
+func (params autoTagParams) get(key paramsKey) string {
+	return params[key]
+}
+
+func (params autoTagParams) set(key paramsKey, value string) {
+	params[key] = value
+}
+
 func (g *AutoParameterGenerator) Source() definition.Source { return definition.Auto }
 func (g *AutoParameterGenerator) Validate(name string, defaultValue interface{}, target reflect.Type) error {
 	err := assignable(defaultValue, target)
@@ -367,7 +382,7 @@ func (g *AutoParameterGenerator) Validate(name string, defaultValue interface{},
 		return invalidAutoParameter.Error(target)
 	}
 	f := func(index []int, field reflect.StructField) error {
-		source, name, err := g.split(field.Tag.Get("source"))
+		source, name, params, err := g.split(field.Tag.Get("source"))
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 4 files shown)


service/source.go
```diff
@@ -385,15 +385,18 @@ func (g *AutoParameterGenerator) Validate(name string, defaultValue interface{},
 	return err
 }
 
-func (g *AutoParameterGenerator) split(tag string) (source string, name string, err error) {
+func (g *AutoParameterGenerator) split(tag string) (source, name, defaultValue string, err error) {
 	result := strings.Split(tag, ",")
+
 	if len(result) == 1 {
```
kdada: Use `switch`.

service/source.go
```diff
@@ -385,15 +385,18 @@ func (g *AutoParameterGenerator) Validate(name string, defaultValue interface{},
...
-		return result[0], "", nil
-	}
-	if len(result) != 2 {
-		return "", "", invalidFieldTag.Error(tag)
+		return strings.TrimSpace(result[0]), "", "", nil
+	} else if len(result) == 2 {
+		return strings.TrimSpace(result[0]), strings.TrimSpace(result[1]), "", nil
+	} else if len(result) == 3 {
+		return strings.TrimSpace(result[0]), strings.TrimSpace(result[1]), strings.TrimSpace(result[2]), nil
```
kdada: Why don't you use the format: `source,name,default=defaultValue` ?

service/source.go
```diff
@@ -427,6 +430,23 @@ func (g *AutoParameterGenerator) generate(ctx context.Context, vc ValueContainer
 		if err != nil {
 			return err
 		}
+		if ins == nil {
+			var value interface{}
+			if defaultValue != "" {
+				var err error
+				if c := ConverterFor(field.Type); c != nil {
```
kdada: What if `c == nil`?

service/source.go
```diff
@@ -448,3 +468,20 @@ func (g *AutoParameterGenerator) enum(index []int, typ reflect.Type, f func(inde
 	}
 	return nil
 }
+
+func zero(t reflect.Type) interface{} {
+	switch t.Kind() {
+	case reflect.Bool:
+		return false
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return 0
```
kdada: It's wrong. You can't treat `int` as `uint`, `uint8` and etc. 
All values should have its own type rather `int`.

service/source.go
```diff
@@ -448,3 +468,20 @@ func (g *AutoParameterGenerator) enum(index []int, typ reflect.Type, f func(inde
...
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return 0
```
kdada: Ditto.

service/source.go
```diff
@@ -448,3 +468,20 @@ func (g *AutoParameterGenerator) enum(index []int, typ reflect.Type, f func(inde
...
+	case reflect.Float32, reflect.Float64:
+		return 0
```
kdada: Ditto.

service/source.go
yejiayu: ins == value(nil)

service/source.go
```diff
@@ -385,15 +385,37 @@ func (g *AutoParameterGenerator) Validate(name string, defaultValue interface{},
...
+	switch len(result) {
+	case 1:
+		source = strings.TrimSpace(result[0])
+	case 2:
+		source = strings.TrimSpace(result[0])
+		name = strings.TrimSpace(result[1])
+	case 3:
+		source = strings.TrimSpace(result[0])
+		name = strings.TrimSpace(result[1])
+		params := strings.Split(strings.TrimSpace(result[2]), ";")
```
kdada: `path,application,default=123,anotherKey=anotherValue` is more preferable than `Path,application,default=123;anotherKey=anotherValue`.

service/executor.go
```diff
@@ -361,8 +361,12 @@ func (e *executor) Execute(ctx context.Context) (err error) {
 		if err != nil {
 			return writeError(ctx, e.producers, err)
 		}
-		if result == nil && p.defaultValue != nil {
-			result = p.defaultValue
+		if result == nil {
+			if p.defaultValue != nil {
+				result = p.defaultValue
+			} else {
+				result = zero(p.targetType)
```
kdada: You must generate a completely correct zero value here.

yejiayu:
ping @kdada

PR #115, opened 2018-01-26
feat(builder): validate operators at startup time

### LINKED ISSUE #73: Throw at startup time instead of run time in case of wrong parameters
**What is the issue about**:
early throw in case of wrong parameters

**What happened**:
if I defined a handler function of type: 
```
func(ctx context.Context, first int, after string)
```

while using definitions as 
```
Parameters: []definition.Parameter{
					{
						Source: definition.Query,
						Name:   "after",
						Type:   reflect.TypeOf(""),
					},
					{
						Source:  definition.Query,
						Name:    "first",
						Type:    reflect.TypeOf(0),
						Default: 10,
					},
				},
```

then there's only an error thrown during request time.

But in fact, this check can be done at startup time.

**Anything else we need to know**:

kdada:
For an operator chain:
```
in -> operators[0].In()
operators[0].Out() -> operators[1].In()
...
operators[N].Out() -> out
```
Parameter types and result types are checked at startup time now.

Fixes #73 

/assign @yejiayu 
/assign @Jimexist 
/area framwork
/kind enhancement
/priority P1

**Release note**:

```release-note
Check parameters and results of operators at startup time
```

### DIFF
--- a/service/executor.go
+++ b/service/executor.go
@@ -50,7 +50,7 @@ func (i *inspector) addDefinition(d definition.Definition) error {
 		return definitionNoMethod.Error(d.Method, i.path)
 	}
 	if len(d.Consumes) <= 0 {
-		return definitionNoComsumes.Error(d.Method, i.path)
+		return definitionNoConsumes.Error(d.Method, i.path)
 	}
 	if len(d.Produces) <= 0 {
 		return definitionNoProduces.Error(d.Method, i.path)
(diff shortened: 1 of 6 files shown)


service/executor_test.go
```diff
@@ -0,0 +1,287 @@
...
+			definition.Definition{
+				Method: definition.Method(""),
+			},
+			definitionNoMethod,
+		},
+		{
+			definition.Definition{
+				Method: definition.Get,
+			},
+			definitionNoComsumes,
```
ddysher: typo: definitionNoConsumes

kdada:
@ddysher PTAL

service/executor.go
```diff
@@ -202,14 +208,23 @@ func (i *inspector) generateResults(funcName string, typ reflect.Type, rs []defi
...
 		}
 		result := result{
 			index:     index,
 			handler:   handler,
 			operators: r.Operators,
 		}
-		if err := handler.Validate(typ.Out(index)); err != nil {
+		outType := typ.Out(index)
+		if len(result.operators) > 0 {
+			LastOperatorOutType := result.operators[len(result.operators)-1].Out()
```
yejiayu: What if `result.operators` > 1?

PR #114, opened 2018-01-26
 example(tracing): Add tracing example

yejiayu:
**What this PR does / why we need it**:
Add tracing example

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:
```release-note
NONE
```

PR #116, opened 2018-01-26
feat(tracing): Add tracing http transport

yejiayu:
Add tracing http transport

Fixes #

**Release note**:

```release-note
NONE
```


plugins/tracing/client.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/opentracing/opentracing-go"
+)
+
+// HTTPClient wraps an http.Client with tracing instrumentation.
+type HTTPClient interface {
+	Do(ctx context.Context, req *http.Request, operationName string) (*http.Response, error)
+}
+
+func NewHTTPClient() HTTPClient {
+	client := &http.Client{Transport: &nethttp.Transport{}}
```
walktall: Maybe a http.Client singleton is better.

kdada:
`tracing.NewRequest(ctx context.Context, xxxxx)`  or an implementation of `tracing.RoundTripper` is more preferable.

plugins/tracing/http.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// RoundTrip implements the RoundTripper interface.
+func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	rt := t.RoundTripper
+	if rt == nil {
+		rt = http.DefaultTransport
+	}
+
+	span, ctx := StartSpanFromContext(req.Context(), "HTTP: "+req.Method)
```
ddysher: `StartSpanFromContext` will return nil if req.Context() doesn't have span, what if I want to start a new span using the transport?

ddysher: The operation name `"HTTP: "+req.Method` doesn't make any sense.

plugins/tracing/http.go
```diff
@@ -0,0 +1,179 @@
...
+	if span == nil {
+		return rt.RoundTrip(req)
+	}
+	defer span.Finish()
+
+	ht := tracer{span: span}
+	req = req.WithContext(httptrace.WithClientTrace(ctx, ht.clientTrace()))
```
ddysher: I'm hesitate to add this.  We should provide an option to enable/disable this for debugging, and http trace should be off by default.

plugins/tracing/http.go
yejiayu: If you want this span to be more meaningful, you should create a span of yourself.

plugins/tracing/http.go
yejiayu: So why are you using `tracing http transport`?

plugins/tracing/http.go
yejiayu: Maybe we should use `HTTP: / API/v1 / applications / : app`  as the operation name?

ddysher: You can have users pass operation name as part of transport construction, and use `HTTP: /api/v1/applications/:app` as a default fallback if user doesn't pass any.

Using `/api/v1/applications/:app` is better since i'm able to tell who is sending the request.  `HTTP: GET`, `HTTP: POST` is REALLY non-sense.

plugins/tracing/http.go
ddysher: I mean enable/disable httptrace, not the whole tracing.  By default, we only trace http transaction, but user is able to turn on httptrace to debug (e.g. DNS issue) if needed.

It's debatable if we want to enable dns hook by default, but please do not turn on all httptrace hooks.

plugins/tracing/http.go
yejiayu: Let me think

yejiayu:
ping @ddysher

kdada:
How to use it? Could you give an example?

yejiayu:
After this PR merge you can see https://github.com/caicloud/nirvana/pull/114

ddysher:
Please add tests

plugins/tracing/http.go
```diff
@@ -0,0 +1,190 @@
...
+// RoundTrip implements the RoundTripper interface.
+func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	rt := t.RoundTripper
+	if rt == nil {
+		rt = http.DefaultTransport
+	}
+
+	operationName := t.OperationName
+	if operationName == "" {
+		operationName = req.Method
```
ddysher: why not adding URL here?

ddysher:
@yejiayu 文件改一下名字，http.go -> client.go, tracing.go -> server.go

plugins/tracing/client.go
```diff
@@ -0,0 +1,190 @@
...
+	if rt == nil {
+		rt = http.DefaultTransport
+	}
+
+	operationName := t.OperationName
+	if operationName == "" {
+		operationName = req.Method
+	}
+
+	span, ctx := StartSpanFromContext(req.Context(), "HTTP: "+operationName)
```
ddysher: my understanding is that roundtrip  will start a new span if there is no span found in the context?

yejiayu:
PTAL @ddysher

plugins/tracing/client.go
ddysher: discussed offline, we don't use global tracer, so it's not necessary

ddysher:
@yejiayu test is failing

yejiayu:
ping @ddysher

plugins/tracing/http.go
ddysher: note to myself: this might not be needed since we added url tag.

PR #134, opened 2018-02-04
fix(openapi): use existing spec path to avoid overwrite path

ddysher:

If existing spec path is found, do not init a new one.

Found the bug when writing readme; If we define multiple methods for the same url, only the last one exists in final api spec, all others will be overwritten.

/cc @liubog2008 

**Release note**:

```release-note
NONE
```


ddysher:
@liubog2008 can you add more unit test in the package (also covers this bug)?

liubog2008:
There is a fatal bug... I'll fix it today = =
This PR can't fix the real problem

liubog2008:
see #137

ddysher:
SGTM

/close

PR #159, opened 2018-02-07
 feat(config): add nirvana config

kdada:

- chore(vendor): update vendor
- feat(config): add nirvana command
- feat(plugins): add plugin options

Fixes #

/assign @ddysher 

**Release note**:

```release-note
Add nirvana config
Add nirvana command
Add plugin options
```

### DIFF
--- a/config/center.go
+++ b/config/center.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2018 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/spf13/cast"
+	"github.com/spf13/viper"
+)
+
+var v = viper.New()
+
+func init() {
+	// Initialize config paths.
+	// These paths are searched:
+	//   ./
+	//   ./config/
+	//   {ExecutableFilePath}/
+	//   {ExecutableFilePath}/config/
+	//   /etc/nirvana/
+	//
+	// Config file should be one of:
+	//   nirvana.yaml
+	//   nirvana.toml
+	//   nirvana.json
+	v.SetConfigName("nirvana")
+	file, err := os.Executable()
+	if err != nil {
+		panic(err)
+	}
+	path := filepath.Dir(file)
+	v.AddConfigPath("./")
+	v.AddConfigPath("./config/")
+	v.AddConfigPath(path + "/")
+	v.AddConfigPath(path + "/config/")
+	v.AddConfigPath("/etc/nirvana/")
+
+	// Read configs.
+	if err := v.ReadInConfig(); err != nil {
+		// Ignore not found error
+		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
+			panic(err)
+		}
+	}
+}
+
+// IsSet checks to see if the key has been set in any of the data locations.
+// IsSet is case-insensitive for a key
+func IsSet(key string) bool {
+	return v.IsSet(key)
+}
+
+// Set sets the value for the key in the override regiser.
+// Set is case-insensitive for a key.
+// Will be used instead of values obtained via
+// flags, config file, ENV, default, or key/value store.
+func Set(key string, value interface{}) {
+	v.Set(key, value)
+}
+
+// Get can retrieve any value given the key to use.
+// Get is case-insensitive for a key.
+// Get has the behavior of returning the value associated with the first
+// place from where it is set. Viper will check in the following order:
+// override, flag, env, config file, key/value store, default
+//
+// Get returns an interface. For a specific value use one of the Get____ methods.
+func Get(key string) interface{} {
+	return v.Get(key)
+}
+
+// GetBool returns the value associated with the key as a bool.
+func GetBool(key string) bool {
+	return cast.ToBool(v.Get(key))
+}
+
+// GetDuration returns the value associated with the key as a time.Duration.
+func GetDuration(key string) time.Duration {
+	return cast.ToDuration(v.Get(key))
+}
+
+// GetFloat32 returns the value associated with the key as a float32.
+func GetFloat32(key string) float32 {
+	return cast.ToFloat32(v.Get(key))
+}
+
+// GetFloat64 returns the value associated with the key as a float64.
+func GetFloat64(key string) float64 {
+	return cast.ToFloat64(v.Get(key))
+}
+
+// GetInt returns the value associated with the key as a int.
+func GetInt(key string) int {
+	return cast.ToInt(v.Get(key))
+}
+
+// GetInt8 returns the value associated with the key as a int.
+func GetInt8(key string) int8 {
+	return cast.ToInt8(v.Get(key))
+}
+
+// GetInt16 returns the value associated with the key as a int.
+func GetInt16(key string) int16 {
+	return cast.ToInt16(v.Get(key))
+}
+
+// GetInt32 returns the value associated with the key as a int32.
+func GetInt32(key string) int32 {
+	return cast.ToInt32(v.Get(key))
+}
+
+// GetInt64 returns the value associated with the key as a int64.
+func GetInt64(key string) int64 {
+	return cast.ToInt64(v.Get(key))
+}
+
+// GetString returns the value associated with the key as a string.
+func GetString(key string) string {
+	return cast.ToString(v.Get(key))
+}
+
+// GetStringSlice returns the value associated with the key as a []string.
+func GetStringSlice(key string) []string {
+	return cast.ToStringSlice(v.Get(key))
+}
+
+// GetUint returns the value associated with the key as a uint.
+func GetUint(key string) uint {
+	return cast.ToUint(v.Get(key))
+}
+
+// GetUint8 returns the value associated with the key as a uint.
+func GetUint8(key string) uint8 {
+	return cast.ToUint8(v.Get(key))
+}
+
+// GetUint16 returns the value associated with the key as a uint.
+func GetUint16(key string) uint16 {
+	return cast.ToUint16(v.Get(key))
+}
+
+// GetUint32 returns the value associated with the key as a uint32.
+func GetUint32(key string) uint32 {
+	return cast.ToUint32(v.Get(key))
+}
+
+// GetUint64 returns the value associated with the key as a uint64.
+func GetUint64(key string) uint64 {
+	return cast.ToUint64(v.Get(key))
+}
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,459 @@
+/*
+Copyright 2018 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/caicloud/nirvana"
+	"github.com/caicloud/nirvana/definition"
+	"github.com/caicloud/nirvana/errors"
+	"github.com/spf13/cast"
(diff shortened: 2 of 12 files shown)


kdada:
Example:
```go
func main() {
	starter := config.NewNirvanaStarter()
	starter.EnablePlugin(
		&nirvana.Plugin{Port: 8080},
		&metrics.Plugin{Path: "/metrics"},
		&profiling.Plugin{Path: "/profiling"},
	)
	if err := starter.Execute(echo); err != nil {
		log.Fatal(err)
	}
}
```

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+// Plugin is for plugins to collect configurations
+type Plugin interface {
+	// Name returns plugin name.
+	Name() string
+	// Configure configures nirvana config via current options.
+	Configure(cfg *nirvana.Config) error
+}
+
+// NirvanaStarter is a nirvana starter.
+type NirvanaStarter interface {
```
zoumo: NirvanaStarter is 😑，maybe NirvanaLauncher looks better.

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+	// EnablePlugin enables plugins.
+	EnablePlugin(plugins ...Plugin) NirvanaStarter
+	// AddOption will fill up options from config/ENV/flags after executing.
+	AddOption(prefix string, options ...Option) NirvanaStarter
+	// Add adds a field by key
+	Add(pointer interface{}, key string, shortFlag string, desc string) NirvanaStarter
```
zoumo: `Add` confuse me, add what? It SHOULDE be AddFlag
and the first arg pointer is not required.
should we provide `AddFlag` and `AddFlagP` just like cobra?

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+		walkthrough([]int{}, typ, func(index []int, field reflect.StructField) {
+			ptr := val.FieldByIndex(index).Addr().Interface()
+			s.Add(ptr, prefix+field.Name, "", field.Tag.Get("desc"))
+		})
+	}
+	return s
+}
+
+// Add adds a field by key
+func (s *starter) Add(pointer interface{}, key string, shortFlag string, desc string) NirvanaStarter {
```
zoumo: add unittest for it

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+// ExecuteWithConfig runs nirvana server from a custom config.
+func (s *starter) ExecuteWithConfig(cfg *nirvana.Config) error {
+	return s.Command(cfg).Execute()
+}
+
+// Command returns a command for starter.
+func (s *starter) Command(cfg *nirvana.Config) *cobra.Command {
+	cmd := &cobra.Command{
+		Use: s.name,
+		Run: func(cmd *cobra.Command, args []string) {
```
zoumo: add Short and Long Usage

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+			if err := nirvana.NewServer(cfg).Serve(); err != nil {
+				cfg.Logger.Fatal(err)
+			}
+		},
+	}
+	fs := cmd.Flags()
+	for _, f := range s.fields {
+		var value, envValue interface{} = nil, nil
+		env := os.Getenv(f.env)
+		switch v := f.pointer.(type) {
```
zoumo: check pointer == nil before type assertion

config/config.go
zoumo: How can I add a persistent flag?

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+		case *uint8:
+			if IsSet(f.key) {
+				value = GetUint8(f.key)
+			}
+			if env != "" {
+				envValue = cast.ToUint8(env)
+			}
+			desc, val := chooseValue(f.key, value, f.env, envValue, *v)
+			fs.Uint8VarP(v, f.longFlag, f.shortFlag, val.(uint8), f.description+desc)
```
zoumo: can you explain the priorities of different config source: 
- explicit calls to set
- flag
- env
- default value
- config file

config/config.go
ddysher: This is used to add to internal store?

kdada: `Add` needs more comments and it SHOULD NOT be `AddFlag`.

`Add` add a field to receive configuration from `Flag/ENV/ConfigFile`.  That means the key is like `nirvana.ip` instead of `nirvana-ip`. Same approach to get configuration is `config.GetString("nirvana.ip")`.

Also, you should not add a persistent flag.

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+        _| |_\   |_  | |  | |     \ \/ / // | |, | | | | // | |, 
+       |_____|\____|[___][___]     \__/  \'-;__/[___||__]\'-;__/ 
+
+`
+
+// Option must be a pointer to struct.
+type Option interface{}
+
+// Plugin is for plugins to collect configurations
+type Plugin interface {
```
ddysher: why is this called `Plugin` instead of `Option`?

This is weird for nirvana since nirvana options will becomes type `Nirvana.Plugin`

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+       s++sys+//yysssso+oysssm///+++////::::::::::::/s:::/+:::/y'''      
+       :+o+++oo/+ssssoosyyhyod:::::::::::::::::::::+o:::+h++:/s'         
+        '-+++/////+++shyyooooys::::::::::::::::::::s++++sooh+s'          
+           './oysoo///osssssssho::::::::::::::::::::/:::o+/so'           
+              ss+++//////////+oys/:::::::::::::::::::::::+o:             
+              '.:/+++++++++//-.'.soo/o::::::::::::::::/oo:'              
+                   ''''''        h/-./ss+++///////++oo/-'                
+                                 oo++/.''o+/y////::-.                    
+                                         ://-                            
+        ____  _____   _
```
caitong93: I prefer not to print the logo

config/config.go
kdada: Flag > ENV > Config file > Default

plugins/metrics/metrics.go
```diff
@@ -187,3 +187,25 @@ func Namespace(ns string) nirvana.Configurer {
 		return nil
 	}
 }
+
+// Plugin contains basic configurations of metrics.
+type Plugin struct {
+	// Namespace is metrics namespace.
+	Namespace string `desc:"Metrics namespace"`
+	// Path is metrics path.
+	Path string `desc:"Metrics handler path"`
```
caitong93: just `Metrics path`  is fine

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+	desc := ""
+	if value != nil {
+		val = value
+		desc += fmt.Sprintf(" (%s=%v)", key, value)
+	} else {
+		desc += fmt.Sprintf(" (%s)", key)
+	}
+	if envValue != nil {
+		val = envValue
+		desc = fmt.Sprintf(" (%s=%v)", env, envValue) + desc
```
ddysher: nit

```
desc = fmt.Sprintf(" (ENV: %s=%v)", env, envValue) + desc
```

make it clear what this is. ditto for others

plugins/metrics/metrics.go
```diff
@@ -187,3 +187,25 @@ func Namespace(ns string) nirvana.Configurer {
 		return nil
 	}
 }
+
+// Plugin contains basic configurations of metrics.
+type Plugin struct {
+	// Namespace is metrics namespace.
+	Namespace string `desc:"Metrics namespace"`
```
ddysher: how about default value?

config/config.go
ddysher: why not?  it's cute :)

ddysher:
At high level, we are mixing plugin and nirvana config.  What's the point of using `nirvana.Plugin`, can I paass it to NewNirvanaStarte?

```go
func main() {
	starter := config.NewNirvanaStarter()
	starter.EnablePlugin(
		&nirvana.Plugin{Port: 8080},
		&metrics.Plugin{Path: "/metrics"},
		&profiling.Plugin{Path: "/profiling"},
	)
	if err := starter.Execute(echo); err != nil {
		log.Fatal(err)
	}
}
```

config/config.go
kdada: I think so. Option is better.

config/center.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/spf13/cast"
+	"github.com/spf13/viper"
+)
+
+var v = viper.New()
+
+func init() {
+	// Initialize config paths.
+	// These paths are searched:
```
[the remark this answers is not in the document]
kdada: Default search path is enough.

config/config.go
```diff
@@ -0,0 +1,459 @@
...
+
+type starter struct {
+	name    string
+	plugins []Plugin
+	fields  map[string]*configField
+}
+
+// EnablePlugin enables plugins.
+func (s *starter) EnablePlugin(plugins ...Plugin) NirvanaStarter {
+	s.plugins = append(s.plugins, plugins...)
```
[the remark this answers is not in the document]
kdada: Yes.

plugins/metrics/metrics.go
kdada: It uses `Namespace` configurer. Defaults to "nirvana".
```go
// Namespace returns a configure to set metrics namespace.
func Namespace(ns string) nirvana.Configurer {
	if ns == "" {
		ns = "nirvana"
	}
	return func(c *nirvana.Config) error {
		wapper(c, func(c *config) {
			c.namespace = ns
		})
		return nil
	}
}
```

kdada:
@ddysher 
May pass nirvana options via parameters is better:
```go
starter := config.NewNirvanaStarter(ip string, port uint16)
```

ddysher:
Will we ever have a lot options?

config/config.go
kdada: Checked in `Add`.

plugins/metrics/metrics.go
ddysher: So you depend on plugin author to know and check input string and provide a default, which is not a user-friendly API.  Since you already provide `desc` as part of option tag, having a `default` tag sounds more natural.

config/config.go
```diff
@@ -0,0 +1,519 @@
...
+	// Execute runs nirvana server.
+	Execute(descriptors ...definition.Descriptor) error
+	// ExecuteWithConfig runs nirvana server from a custom config.
+	ExecuteWithConfig(cfg *nirvana.Config) error
+	// Command returns a command for command.
+	Command(cfg *nirvana.Config) *cobra.Command
+}
+
+// NewNirvanaCommand creates a nirvana command.
+func NewNirvanaCommand(option NirvanaOption) NirvanaCommand {
```
ddysher: Can you use pointer here, so it will be consistent with passing options to plugins.

ddysher:
@kdada 

High-level programming interface almost LGTM, to summarize my comments

- add a `default` tag for options
- pass "pointer to nirvana config" in NewNirvanaCommand argument to be more consistent with other plugins
- since config will be the entry point, can we move nirvana.go to its own package? a larger problem that I see is there is a mix of struct like config.Option, config.NirvanaOption, nirvana.Config, all of them are public struct, which can be really confusing to users

Once this is done, please add unittest, and update example in other PRs

ddysher:
@zoumo please take another look

kdada:
@ddysher 

> add a default tag for options

Tag `default` is useless. Default value is from instance directly.

> pass "pointer to nirvana config" in NewNirvanaCommand argument to be more consistent with other plugins

It's pointer now.

> since config will be the entry point, can we move nirvana.go to its own package? a larger problem that I see is there is a mix of struct like config.Option, config.NirvanaOption, nirvana.Config, all of them are public struct, which can be really confusing to users

I renamed `config.Option` to `config.CustomOption`, also `config.NirvanaOption` to `config.Option`. Now we only have two concepts: `nirvana.Config` and `Option`.
Fields in `Option`s are filled by `config.NirvanaCommand` after executing.

```go
func main() {
	cmd := config.NewDefaultNirvanaCommand()
	cmd.EnablePlugin(
		&metrics.Option{Path: "/metrics"},
		&profiling.Option{Path: "/profiling"},
	)
	if err := cmd.Execute(echo); err != nil {
		log.Fatal(err)
	}
}
```

ddysher:
LGTM

@zoumo

config/config.go
```diff
@@ -445,15 +505,15 @@ func chooseValue(key string, value interface{}, env string, envValue interface{}
 	desc := ""
 	if value != nil {
 		val = value
-		desc += fmt.Sprintf(" (%s=%v)", key, value)
+		desc += fmt.Sprintf(" (CFG: %s=%v)", key, value)
```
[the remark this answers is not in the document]
zoumo: since `default` is lowercase

config/config.go
```diff
@@ -445,15 +505,15 @@ func chooseValue(key string, value interface{}, env string, envValue interface{}
...
 	} else {
-		desc += fmt.Sprintf(" (%s)", key)
+		desc += fmt.Sprintf(" (CFG: %s)", key)
 	}
 	if envValue != nil {
 		val = envValue
-		desc = fmt.Sprintf(" (%s=%v)", env, envValue) + desc
+		desc = fmt.Sprintf(" (ENV: %s=%v)", env, envValue) + desc
```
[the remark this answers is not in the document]
zoumo: ditto

kdada:
@ddysher  @zoumo  PTAL

ddysher:
/lgtm

??

PR #165, opened 2018-02-09
fix(definition): remove builder pattern

### LINKED ISSUE #125: some feedback while using definition API

/kind feature

While using the definition API I found that it is _feature rich_, i.e. there are many ways to define a service endpoint:

### by declaratively construction:

```go
definition.Descriptor{
	Path:     "/version",
	Produces: []string{definition.MIMEJSON},
	Definitions: []definition.Definition{
		{
			Method:      definition.Get,
			Description: "returns version information about this service",
			Results: []definition.Result{
				{
					Description: "version info",
					Destination: definition.Data,
				},
			},
			Function: func(_ context.Context) Version {
				return Version{
					Name:    name,
					Version: version.Version,
					Commit:  version.Commit,
				}
			},
		},
	},
}
```

### by using builder patterns:

```go
definition.DescriptorFor("/version", "version information",
).Produce(definition.MIMEJSON,
).Definition(definition.GetDefinitionFor(Version{}, "returns version information",
).Result(definition.DataResultFor("version info"),
))
```

_note that 👆 I couldn't find place to define the actual function_

But the problem is I think a simple API should be simple enough for users to _not_ choose, or at least not get confused on mismatching two different sets of apis. A good starting point is to separate them into two different packages.

But in long term this is still a problem because having two sets of API means you have to have _two sets of docs_ and _a third one explaining pros and cons and which one to use in which situation_. I think it unnecessarily complicates the still immature framework. 

Have a simple and unified API, docs and style and them iterate on them.

**Sometimes, less is significantly more, esp. the case with APIs. You can always add features, but it is crazily hard to remove**.

kdada:

- fix(definition): remove builder pattern

Fixes #125 

/assign @Jimexist 
/assign @ddysher 
/kind cleanup

**Release note**:

```release-note
Remove builder pattern for definition
```


Jimexist:
does it even make sense to have something like convert from v1 to v2, since we are at very early stage of this framework and there's not too many projects using this just yet.

ddysher:
@kdada LGTM, but please add config test first :)

kdada:
@Jimexist 
Converter is a kind of operator. We don't need it in the beginning.

PR #181, opened 2018-03-04
refactor(tracing): change dir

yejiayu:

Add your description

Fixes #

/cc @ddysher @kdada 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/examples/tracing/main.go
+++ b/examples/tracing/main.go
@@ -24,7 +24,7 @@ import (
 	"github.com/caicloud/nirvana/config"
 	"github.com/caicloud/nirvana/definition"
 	"github.com/caicloud/nirvana/log"
-	"github.com/caicloud/nirvana/plugins/tracing"
+	tracingMiddleware "github.com/caicloud/nirvana/plugins/tracing/middleware"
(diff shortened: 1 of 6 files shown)


ddysher:
@yejiayu test is failing

plugins/tracing/client/http.go
```diff
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package tracing
+package client
```
ddysher: will this become a problem when we add more clients?  e.g. name conflict, client.Transport is required for both http and other protocol.

/cc @kdada

yejiayu: We can use the protocol name as a prefix

kdada: `plugins/tracing/clients/http/client.go` is more flexible.

ddysher: > plugins/tracing/clients/http/client.go is more flexible.

SGTM

yejiayu:
ping @ddysher

PR #186, opened 2018-03-14
fix(service): treat Content-Length "" and "0" equally

zoumo:

Add your description

Fixes #

/cc @kdada

**Release note**:

```release-note
Treat Content-Length "" and "0" equally
```

### DIFF
--- a/service/filter.go
+++ b/service/filter.go
@@ -86,15 +86,15 @@ func ParseRequestForm() Filter {
 	}
 }
 
-func isZero(length string) bool {
+func isGTZero(length string) bool {
 	if length == "" {
-		return true
+		return false
 	}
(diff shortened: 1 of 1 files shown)


service/filter.go
```diff
@@ -86,13 +86,24 @@ func ParseRequestForm() Filter {
...
 
+func isZero(length string) bool {
+	if length == "" {
+		return true
+	}
+	i, err := strconv.Atoi(length)
+	if err != nil {
+		return false
+	}
+	return i == 0
```
kdada: The `i` may be negative.
You can merge the two conditions with:
```go
return err != nil || i <= 0
```

zoumo: i == 0 should return true...

PR #190, opened 2018-03-23
feat(config): add PreConfigure and PostConfigure hooks

kdada:

Add PreConfigure and PostConfigure hooks.
After creating nirvana server, config becomes immutable. So we need the two hooks.

Fixes #

/cc @caitong93 
/cc @zoumo 

**Release note**:

```release-note
Add PreConfigure and PostConfigure hooks
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -326,8 +326,10 @@ func (s *command) Add(pointer interface{}, key string, shortFlag string, desc st
 // Execute runs nirvana server.
 func (s *command) Execute(descriptors ...definition.Descriptor) error {
 	cfg := nirvana.NewDefaultConfig()
-	cfg.Configure(nirvana.Descriptor(descriptors...))
-	return s.Command(cfg).Execute()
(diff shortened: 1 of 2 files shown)


ddysher:
what's the use case?

kdada:
@ddysher 
Users want to install some middlewares or handlers into config. But once `nirvana.NewServer` is called, its config will be locked and can't be modified.

caitong93:
This is complicated, for more flexibility maybe we should let user to write `Run` function rather than generated it.
```
type CommandContext interface {
   Configure() error
}
cmd.Command(func(cc CommandContext, server server.Server) error {
   if err := cc.Configure(); err != nil {
       return err
    }
    //  Bootstrap: connect database, etc.
    //  Start the server.
    //  Shutdown the server.
    ...
})

```

cc @kdada

kdada:
There is not only a `Run` function and you can't configure after server is created. So the `NirvanaCommandHook` provides four check points to users.

PR #197, opened 2018-04-02
 feat(service): support ErrorProduces

### LINKED ISSUE #169: Need to distinguish success produces and error produces

 /kind feature

**What happened**:

There is only one field named `Produces` to decide data type of response body.
It's normal for `JSON` and `XML`. Because they have consistent behavior between success data and error data. 
But for those APIs that return files or streams, success data is file reader and error data is `JSON`. In these occasions, We need a mechanism to describe it.

We can simply add a field `ErrorProduces` into `definition.Definition`.
If the field is empty, framework can treat it as `Produces`.

kdada:

Add `ErrorProduces` into `Definition`. `ErrorProduces` is used to generate data for error. If this field is empty, it stands for this field equals to Produces. In some cases, succeessful data and error data should be generated in different ways.

e.g.
An API  returns a file (`Content-Type` is `application/java-archive`), but it may returns an error with `NotFound`. You can set `Definition` as:
```go
Produces: []string{"application/java-archive"},
ErrorProduces: []string{"application/json"},
```
If a file existing, the API returns a file with `Content-Type: application/java-archive`, and carries data by response body. If an error occurs, it returns an error json with `Content-Type: application/json`.

Fixes #169 

/cc @ddysher 

**Release note**:

```release-note
NONE
```


service/utils.go
```diff
@@ -55,6 +55,8 @@ var noConnectionHijacker = errors.InternalServerError.Build("Nirvana:Service:NoC
 var definitionNoMethod = errors.InternalServerError.Build("Nirvana:Service:DefinitionNoMethod", "no http method in [${method}]${path}")
 var definitionNoConsumes = errors.InternalServerError.Build("Nirvana:Service:DefinitionNoConsumes", "no content type to consume in [${method}]${path}")
 var definitionNoProduces = errors.InternalServerError.Build("Nirvana:Service:DefinitionNoProduces", "no content type to produce in [${method}]${path}")
+var definitionNoErrorProduces = errors.InternalServerError.Build("Nirvana:Service:DefinitionNoErrorProduces",
```
zjj2wry: should be var(
     definitionNoErrorProduces = errors.InternalServerError.Build("Nirvana:Service:DefinitionNoErrorProduces",
)

zjj2wry:
I have a problem if I set the return type of json type, but returned text in handle. This will find the problem at runtime or when compiling?

kdada:
@zjj2wry 
There are some rules:
1. If you returns []byte or string, there is no check for your data.
2. If you returns a struct, it will be converted to target data type (here is json).

ddysher:
@kdada there's conflict

service/builder.go
```diff
@@ -176,7 +179,9 @@ func (b *builder) Build() (Service, error) {
 			for _, d := range bd.definitions {
 				b.logger.V(log.LevelDebug).Infof("  Method: %s Consumes: %v Produces: %v",
 					d.Method, d.Consumes, d.Produces)
-				b.modifier(&d)
+				if b.modifier != nil {
```
ddysher: is this relevant for this PR?

service/utils.go
```diff
@@ -55,6 +55,8 @@ var noConnectionHijacker = errors.InternalServerError.Build("Nirvana:Service:NoC
...
+	"no content type to error produce in [${method}]${path}")
```
ddysher: nit: no content type to produce error in [${method}]${path}

ddysher:
curious how other framework handles such case? i don't recall explicitly specifying error producer while using other frameworks.

service/builder.go
kdada: No. It fixes a little issue.

kdada:
Other frameworks let yourself to judge how to write data back to response. But we have different behaviors to handle APIs. Producers must know which content type is desired. In most time, success data and error data have same content type, then a `Produces` is enough. But in some cases, e.g. download APIs,  They need a certain type for files and another type for errors.

ddysher:
> Other frameworks let yourself to judge how to write data back to response. But we have different behaviors to handle APIs. Producers must know which content type is desired. In most time, success data and error data have same content type, then a Produces is enough. But in some cases, e.g. download APIs, They need a type for files and another type for errors.

sgtm

PR #209, opened 2018-05-11
feat(command): add nirvana basic command

kdada:

This pr implement two commands: `init` and `api`.

```
$ nirvana init /path/to/your/project
```
The `init` command initializes a common project structure for you.

```
$ nirvana api /path/to/yout/project/pkg/api
```
The `api` command hosts a server to present api docs for you.

Fixes #

/cc @ddysher 

**Release note**:

```release-note
Add nirvana basic command
```


kdada:
Review Guide:

The following two commits remove old openapi generator and update vendor. Don't need to review.
1. chore(vendor): update vendor
2. chore(openapi): remove deprecated openapi

The following five commits have less codes. You can have a quick review.
1. add default option creators
2. move banner to root directory
3. exit when fatal in silent logger
4. add summary to definition
5. refactor for more usability

This commit implements `init` command (`main.go` is in the next commit). It's used to create project. 
1. add init to nirvana command

This commit implements a basic middle layer for api and clients (not implement yet). Also there is a `api` command to generate api docs and host a server to view docs in browser. 
1. feat(api): support api generation and serving

Above codes uses some tricks, so I ignore some checks here.
1. chore(lint): ignore some checks


Total additions are about 3000 lines.

ddysher:
Superb!

A few comments on generated structure:

- Add a `README.md` file with short introductions about the generated project (the purpose of this readme file is to guide users to create their own README file)
- Add `docs.go` for each package
- How to handle `pkg/database`, as discussed offline?

kdada:
`README.md` and `docs.go`, we can add them in future.

Nirvana is a communication layer for services and independent with logic. It does not care about data layer. But we can provide a common practice in our future docs.

service/builder.go
```diff
@@ -189,16 +255,16 @@ func (b *builder) Build() (Service, error) {
 
 			leaf.SetInspector(inspector)
 		}
-		if len(bd.middlewares) > 0 {
-			leaf.AddMiddleware(bd.middlewares...)
+		for _, m := range bd.middlewares {
```
ddysher: looks like this is fixing a bug?

service/source.go
```diff
@@ -510,3 +462,59 @@ func (g *AutoParameterGenerator) enum(index []int, typ reflect.Type, f func(inde
...
+
+// Set sets value for a config key.
+func (f AutoParameterConfig) Set(key AutoParameterConfigKey, value string) {
+	f[key] = value
+}
+
+// ParseAutoParameterTag parse the tag of AutoParameter.
+func ParseAutoParameterTag(tag string) (source definition.Source, name string, apc AutoParameterConfig, err error) {
+	apc = make(AutoParameterConfig)
+	// This operation is problematic.
```
ddysher: this method hard-code quite a few things, which are actually user-facing (part of the framework API)?

cmd/nirvana/project/init.go
```diff
@@ -0,0 +1,782 @@
...
+│   │       ├── descriptors
+│   │       │   ├── descriptors.go
+│   │       │   └── message.go
+│   │       └── middlewares
+│   │           └── middlewares.go
+│   ├── message
+│   │   └── message.go
+│   └── version
+│       └── version.go
+├── project.yaml
```
ddysher: `project.yaml` is a bit confusing, how about just `config.yaml` or `nirvana.yaml`?

cmd/nirvana/project/init.go
```diff
@@ -0,0 +1,782 @@
...
+- name: v1
+  description: The v1 version is the first version of this project
+  rules:
+  - "^/api/v1.*"
+`
+}
+
+func (o *initOptions) Manuals() string {
+	return `
+This command generates standard nirvana project structure.
```
ddysher: As mentioned in another comment, please include `README.md` and `doc.go`.

cmd/nirvana/project/init.go
```diff
@@ -0,0 +1,782 @@
...
+{{ .SharpBoilerplate }}
+#
+{{ end -}}
+# Gopkg.toml example
+#
+# Refer to https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md
+# for detailed Gopkg.toml documentation.
+#
+# required = ["github.com/user/thing/cmd/thing"]
+# ignored = ["github.com/user/project/pkgX", "bitbucket.org/user/project/pkgA/pkgY"]
```
ddysher: what's this .... do you need them?

cmd/nirvana/project/init.go
```diff
@@ -0,0 +1,782 @@
...
+
+package descriptors
+
+import (
+	"{{ .ProjectPackage }}/pkg/message"
+
+	def "github.com/caicloud/nirvana/definition"
+)
+
+func init() {
```
ddysher: I'm a little concerned about using init() to register handlers, is there a better alternative?

plugins/tracing/tracing.go
```diff
@@ -60,6 +60,7 @@ func (i *tracingInstaller) Install(builder service.Builder, cfg *nirvana.Config)
 	wapper(cfg, func(c *config) {
 		if c.tracer == nil {
 			tcfg := tconfig.Configuration{
+				ServiceName: c.serviceName,
```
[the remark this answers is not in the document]
kdada: We can add those configs to `Option` later.

log/silent.go
```diff
@@ -63,15 +65,21 @@ func (*SilentLogger) Errorln(...interface{}) {}
 
 // Fatal logs to the FATAL logs, then calls os.Exit(1).
 // Arguments are handled in the manner of fmt.Print; a newline is appended if missing.
-func (*SilentLogger) Fatal(v ...interface{}) {}
+func (*SilentLogger) Fatal(v ...interface{}) {
+	os.Exit(1)
```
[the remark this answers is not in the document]
kdada: golang standard library `log`:
```go
// Fatal is equivalent to Print() followed by a call to os.Exit(1).
func Fatal(v ...interface{}) {
	std.Output(2, fmt.Sprint(v...))
	os.Exit(1)
}
```

utils/api/analyzer.go
```diff
@@ -0,0 +1,152 @@
+/*
```
ddysher: nit: why do you put all these under `utils`?  utils is for shareable code among different packages, looks like we'd better just put it under `openapi/blabla`?

utils/api/definitions.go
```diff
@@ -0,0 +1,184 @@
...
+	// Description describes the example.
+	Description string
+	// Type is result object type.
+	Type TypeName
+	// Instance is encoded instance data.
+	Instance []byte
+}
+
+// Definition is complete version of def.Definition.
+type Definition struct {
```
ddysher: why do you redefine everything here again?.... this is not maintainable

config.json
```diff
@@ -21,7 +21,12 @@
   "Aggregate": true,
   "WarnUnmatchedNolint": true,
   "LineLength": 160,
-  "Exclude": ["comment or be unexported"],
+  "Exclude": [
```
ddysher: are u sure these are the right things to do....

service/builder.go
```diff
@@ -121,40 +142,85 @@ func (b *builder) addDescriptor(prefix string, consumes []string, produces []str
...
+	copy(newOne.Consumes, consumes)
+
+	if len(d.Produces) > 0 {
+		produces = d.Produces
+	}
+	newOne.Produces = make([]string, len(produces))
+	copy(newOne.Produces, produces)
+
+	if len(d.ErrorProduces) > 0 {
+		produces = d.ErrorProduces
```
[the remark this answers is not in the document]
kdada: It means we use specified `ErrorProduces` rather than inherit from `Produces`.  If you want to use   produces, you can simply leave `ErrorProduces` empty.

service/builder.go
```diff
@@ -121,40 +142,85 @@ func (b *builder) addDescriptor(prefix string, consumes []string, produces []str
...
 	}
+	newOne.ErrorProduces = make([]string, len(produces))
+	copy(newOne.ErrorProduces, produces)
+
 	newOne.Parameters = make([]definition.Parameter, len(d.Parameters))
 	for i, p := range d.Parameters {
-		p.Description = ""
-		newOne.Parameters[i] = p
+		newParameter := p
```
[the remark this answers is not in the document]
kdada: Yes.

service/builder.go
```diff
@@ -121,40 +142,85 @@ func (b *builder) addDescriptor(prefix string, consumes []string, produces []str
...
+		if len(bd.middlewares) > 0 {
+			middlewares := make([]definition.Middleware, len(bd.middlewares))
+			copy(middlewares, bd.middlewares)
+			result[path] = middlewares
+		}
+	}
+	return result
+}
+
+// Definitions returns all definitions. If a modifier exists, it will be executed.
```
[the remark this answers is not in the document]
kdada: `Definitions` copies and returns new definitions every time. So modifiers don't need to be idempotent.

service/builder.go
kdada: No. It's from line 105:
```go
if len(descriptor.Middlewares) > 0 {
	for _, m := range descriptor.Middlewares {
		func(m definition.Middleware) {
			bd.middlewares = append(bd.middlewares, func(ctx context.Context, chain router.RoutingChain) error {
				return m(ctx, chain)
			})
		}(m)
	}
	bd.middlewares = append(bd.middlewares, descriptor.Middlewares...)
}
```

service/source.go
kdada: Yes. This method parses field tags from `Auto` parameter.

utils/api/analyzer.go
kdada: Because this package is not only serving for openapi.

utils/api/definitions.go
kdada: I'll refactor this package when I'm writing code generation.

cmd/nirvana/project/init.go
kdada: It's document for users. Of course, we can remove it.

ddysher:
?

kdada:
@ddysher  
Waiting for review @liubog2008 and @caitong93

cmd/nirvana/project/init.go
```diff
@@ -0,0 +1,782 @@
...
+		"pkg/api/api.go":                            o.templateAPI(),
+		"pkg/message/message.go":                    o.templateMessage(),
+		"pkg/version/version.go":                    o.templateVersion(),
+		"Gopkg.toml":                                o.templateGopkg(),
+		"Makefile":                                  o.templateMakefile(),
+		"project.yaml":                              o.templateProject(),
+	}
+}
+
+func (o *initOptions) templateMain() string {
```
caitong93: It is better to split out these templates to dedicated files for better management. Also we should provide a way to custom the default templates.

kdada:
Summary:
1. Fix comment typos.
2. Rename `project.yaml` to `nirvana.yaml`
3. Add `README.md`

These codes is very ugly, but I'll refactor them in code gen.

cmd/nirvana/project/init.go
caitong93: @kdada

cmd/nirvana/api/api.go
```diff
@@ -0,0 +1,358 @@
...
+
+func init() {
+	err := service.RegisterProducer(service.NewSimpleSerializer("text/html"))
+	if err != nil {
+		log.Fatalln(err)
+	}
+}
+
+func newAPICommand() *cobra.Command {
+	options := &apiOptions{}
```
zjj2wry: flags.StringVar(&o.Serve, "serve", o.Server, "Start a server to host api docs")
flags.StringVar(&o.Output, "output", o.Output, "Directory to output api specifications")

add default value to &apiOptions{},  It's easier to add unit tests

kdada:
@zjj2wry 

The first two questions can be finished in the future. And the third one has fixed in #212. We can record these problems and fix them in other small PRs.

log/silent.go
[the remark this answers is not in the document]
ddysher: hmm, such inconsistency ...

zjj2wry:
> The first two questions can be finished in the future. And the third one has fixed in #212. We can record these problems and fix them in other small PRs.

lgtm

cmd/nirvana/project/init.go
ddysher: :point_up_2: what about this?

ddysher:
Looks like serve is not picking up the right flag?

```
nirvana api pkg/api --serve "localhost:8888"
INFO  0613-10:11:45.614+08 api.go:204 | Listening on :8080. Please open your browser to view api docs
FATAL 0613-10:11:45.614+08 api.go:60 | listen tcp :8080: bind: address already in use
```

- Please do make sure to fix issues mentioned here in future PRs; especially the one mentioned from @zjj2wry which is user facing. Also, i suggest we can make `/path/to/apis` optional in `nirvana api` (set the default to `pkg/api` as we already have a recommended structure.
- PS. let's change `pkg/api` to `pkg/apis`, since we mostly always have more than 1 API in a project :)
- I'll send a patch to update README template shortly.

kdada:
@ddysher  @zjj2wry 


We need a baseline to discuss what the next is. I'll file an issue to record and discuss what we need to do.
For more improvements and minor changes, we can add/modify them in the future.

ddysher:
/lgtm

we'll discuss remaining issues in https://github.com/caicloud/nirvana/issues/213

PR #211, opened 2018-06-06
feat(method): support method in definition

kdada:

In the past, `definition.Function` only support functions. With `method` package in this pr, you can get the power to make a method be like a function.

```go
// Before nirvana server starting
instance := InitSomeClass()
method.Put(instance)

// In definition:
{
    ...
    Function: method.Get(&SomeClass{}, "SomeMethod"),
    ... 
}
```

Then you don't need to do this:
```go
func (* SomeClass) Definitions()[]definition.Definition {
    return []definition.Definition {
        ...
        Function: this.SomeMethod,
        ... 
    }
}
``` 

Fixes #

Reference to #

/cc @ddysher 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/definition/method/method_test.go
+++ b/definition/method/method_test.go
@@ -33,7 +33,16 @@ func (t *TestStruct) Number() int {
 	return t.a
 }
 
+type TestStruct2 struct {
+	a int
+}
+
+func (t *TestStruct2) Number() int {
+	return t.a * 2
+}
+
 func TestContainer(t *testing.T) {
+	var ts TestInterface = &TestStruct2{100}
 	cases := []struct {
 		typ      interface{}
 		instance interface{}
@@ -42,6 +51,8 @@ func TestContainer(t *testing.T) {
 	}{
 		{&TestStruct{}, &TestStruct{100}, "Number", 100},
 		{(*TestInterface)(nil), &TestStruct{100}, "Number", 100},
+		{(*TestInterface)(nil), TestInterface(&TestStruct{100}), "Number", 100},
+		{(*TestInterface)(nil), ts, "Number", 200},
 	}
 	for _, c := range cases {
 		f := Get(c.typ, c.method)
--- a/definition/method/method_test.go
+++ b/definition/method/method_test.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2018 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package method
+
+import (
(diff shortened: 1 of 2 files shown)


ddysher:
@kdada few comments:

- we are constructing a struct instance when registering the function method, what is the implication behind this, i.e.
  - what if user wants to instantiate the struct with field value, e.g. `Function: method.Get(&SomeClass{Some: 'abc'}, "SomeMethod"),`, is this a valid use case?
  - what happens to the object (&SomeClass{}) after registration?
  - what if user wants to create another instance, do we want them to reuse the object?
- you seems to use MakeFunction to wrap the registered function, what is the performance penalty?

It's common to register multiple methods of a class, in current setup, users are supposed to do this multiple times, each time with a new instance of the object?

```
{
    ...
    Function: method.Get(&SomeClass{}, "SomeMethod1"),
    Function: method.Get(&SomeClass{}, "SomeMethod2"),
    Function: method.Get(&SomeClass{}, "SomeMethod2"),
    ... 
}
```

ddysher:
:point_up_2: @kdada can u provide some pprof results (esp. memory for objects)?  that will make me feel better ...

kdada:
@ddysher 

The `method` package provides a pattern to `Get` and `Put` instances.

`Get` returns a lazy loaded function. The function has a certain function prototype and an indefinite execution body. Related instance of the function won't be set until someone calls the function. This mechanism ensure that we can separate API definitions and instances.
The first parameter of `Get` is only for extracting instance type. So a filled object makes no sense.

`Put` is simple. The method just put an instance into method container. But for a type, there only one instance is valid. That means if you `Put` many instances with same type, only the last one is used.


Benchmark:
```
BenchmarkContainer-8     3000000               406 ns/op
BenchmarkDirect-8       10000000               222 ns/op
```
Function warp has twice consumption than direct call. But it isn't a big deal. Because the overhead of an entire API request is in microsecond.  (PS. I'll write an unsafe calling package to skip the safe check of reflect packge. )

definition/method/method_test.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, c := range cases {
+		f := Get(c.typ, c.method)
+		if defaultContainer.typeOf(c.typ).Kind() == reflect.Interface {
+			PutInterface(c.typ, c.instance)
+		} else {
+			Put(c.instance)
+		}
+		result := f.(func() int)()
+		if result != c.result {
+			t.Fatalf("Function result should be %d, but got: %d", c.result, result)
```
zjj2wry: %v

definition/method/method_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type TestStruct struct {
+	a int
+}
+
+func (t *TestStruct) Number() int {
+	return t.a
+}
+
```
zjj2wry: add another struct implement TestInterface，used for test putInterface.

+func (t *Test2Struct) Number() int {
+	return t.a*2
+}

definition/method/method_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestContainer(t *testing.T) {
+	cases := []struct {
+		typ      interface{}
+		instance interface{}
```
zjj2wry: func NewTest(number int) TestInterface{
    return &TestStruct{
    a: number, 
   }
}

Usually the initialization instance is a factory, and the instance can use the following method?

PR #229, opened 2018-07-06
fix(generators): correct conflicting types and fix two generators

kdada:

This PR fixes several bugs about generators.

- Add conflict counter to record how many different types with same pkgPath and Name. (`utils/api/types.go`)
  - This issue is from package `method`. `method` may creates a lot of functions with same pkgPath and Name.
- Correct inline struct generation in swagger (`utils/generators/swagger/generator.go`)
- Correct interface generation in swagger (`utils/generators/swagger/generator.go`)
- Only export type definitions which are from current project. (`utils/generators/golang/helper.go`)
  - Types from standard libraries and third-party libraries are not generated to client packages.

Fixes #

Reference to #

/cc @ddysher 
/cc @zjj2wry 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/utils/api/types.go
+++ b/utils/api/types.go
@@ -97,7 +97,7 @@ func (t *Type) RawTypeName() TypeName {
 	return TypeName(t.PkgPath + "." + t.Name)
 }
 
-// TypeName retruns type unique name.
+// TypeName returns type unique name.
 func (t *Type) TypeName() TypeName {
 	tn := t.RawTypeName()
 	if tn != TypeNameInvalid && t.Conflict > 0 {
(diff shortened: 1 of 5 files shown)


utils/api/types.go
```diff
@@ -94,18 +97,27 @@ func (t *Type) TypeName() TypeName {
 	return TypeName(t.PkgPath + "." + t.Name)
 }
 
+// TypeName retruns type unique name.
```
zjj2wry: return

zjj2wry:
/lgtm
i did not read the logic carefully, but the code lgtm

PR #236, opened 2018-09-05
feat(apidocs): add plugin apidocs

bbbmj:

Add plugin apidocs to return swagger.json

Fixes #

Reference to #

/cc @liubog2008 

**Release note**:

```release-note
Add plugin apidocs
```

### DIFF
--- a/plugins/apidocs/apidocs.go
+++ b/plugins/apidocs/apidocs.go
@@ -1,0 +1,212 @@
+/*
+Copyright 2018 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apidocs
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/caicloud/nirvana"
+	"github.com/caicloud/nirvana/definition"
+	"github.com/caicloud/nirvana/service"
+)
+
+func init() {
+	// Register apidocs config installer into nirvana.
+	nirvana.RegisterConfigInstaller(&apidocsInstaller{})
+}
+
+// ExternalConfigName is the external config name of apidocs.
+const ExternalConfigName = "apidocs"
+
+type apidocsInstaller struct{}
+
+// Name is the external config name.
+func (i *apidocsInstaller) Name() string {
+	return ExternalConfigName
+}
+
+type config struct {
+	path  string
+	files map[string]string
+}
+
+// Install installs stuffs before server starting.
+func (i *apidocsInstaller) Install(builder service.Builder, cfg *nirvana.Config) error {
+	var err error
+	wrapper(cfg, func(c *config) {
+		var descriptors []definition.Descriptor
+		var routerPaths []string
+
+		for path, filePath := range c.files {
+			b, err := ioutil.ReadFile(filePath)
+			if err != nil {
+				panic(err)
+			}
+
+			router := strings.Join([]string{strings.TrimRight(c.path, "/"), strings.Trim(path, "/")}, "/")
+			routerPaths = append(routerPaths, router)
+
+			ds := definition.Descriptor{
+				Path:     path,
+				Consumes: []string{definition.MIMEAll},
+				Produces: []string{definition.MIMEJSON},
+				Definitions: []definition.Definition{{
+					Method:  definition.Get,
+					Results: definition.DataErrorResults(""),
+					Function: func(ctx context.Context) ([]byte, error) {
+						return b, nil
+					}},
+				}}
+
+			descriptors = append(descriptors, []definition.Descriptor{ds}...)
+		}
+
+		err = builder.AddDescriptor(definition.Descriptor{
+			Path:     c.path,
+			Consumes: []string{definition.MIMEAll},
+			Produces: []string{definition.MIMEJSON},
+			Children: descriptors,
+			Definitions: []definition.Definition{{
+				Method:  definition.List,
+				Results: definition.DataErrorResults(""),
+				Function: func(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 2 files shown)


bbbmj:
usage can be:

code in `cmd/myproject/main.go`

```
	swaggerOption := swagger.NewOption(map[string]string{
		"/apis/v1/swagger.json": "api.v1.json",
		"/apis/v2/swagger.json": "api.v2.json",
	})

	// Enable plugins.
	cmd.EnablePlugin(metricsOption, loggerOption, reqlogOption, versionOption, swaggerOption)
```

bbbmj:
As command to generate swagger.json `nirvana api` depends on binary `nirvana`, it's not friendly to try to get binary `nirvana` in Dockerfile. 

How about making more docs to make it up?

plugins/swagger/swagger.go
```diff
@@ -0,0 +1,180 @@
...
+		return "", "", fmt.Errorf("Please specify the swagger config with apiVersion:swaggerFilePath")
+	}
+
+	set := strings.Split(config, ":")
+
+	if len(set) != 2 {
+		return "", "", fmt.Errorf("Please specify the repository name with apiVersion:swaggerFilePath")
+	}
+
+	path := fmt.Sprintf("/apis/%s/swagger.json", strings.TrimSpace(set[0]))
```
kdada: Don't assume that users will use `/apis/`.

bbbmj: updated

ddysher:
How useful is it to just serve swagger.json file?  is it possible that we also serve redoc under path `apis/v1`?

one option is that we update our makefile & dockerfile to use two stage build.  The first stage contains a builder image which includes the `nirvana` binary and outputs `swagger.json`, and the second stage serves the file.  we can provide a new command like `make with-apidoc`?
> As command to generate swagger.json nirvana api depends on binary nirvana, it's not friendly to try to get binary nirvana in Dockerfile.

bbbmj:
> How useful is it to just serve swagger.json file? is it possible that we also serve redoc under path apis/v1?

After discussing with @liubog2008 , we decided to just serve swagger.json file as we don't need many ReDoc services in a cluster.

plugins/swagger/swagger.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil
+	}
+}
+
+type config struct {
+	config []string
+}
+
+// SetOptions returns a configurer to set swagger options.
+func SetOptions(option Option) nirvana.Configurer {
```
kdada: Please refer to https://github.com/caicloud/nirvana/blob/master/plugins/reqlog/reqlog.go.
The plugin does not obey the convention of Nirvana.

Option is a convenient way for generating command flags/ENV. It's not for users to set plugin config.

kdada: Also you need to provide `Default()` and `Disable()`.

kdada:
After examining this use case,  I think you should provide a resource API to cover it.

For example:
`/apidocs` returns a list of the information of all API docs.
`/apidocs/{apidoc}` returns an API doc.

Then users can specify:
1. `--apidocs-path=/apidocs`
2. `--apidocs-files=v1:./api.v1.json,v2:./api.v2.json`

bbbmj:
With this plugin, user can add following to Makefile

```
# Path to your apis
API_DIR ?= pkg/apis

# Api docs output directory.
API_OUTPUT_DIR := ./apis

.PHONY: apidocs

apidocs:
	@nirvana api $(API_DIR) --serve="" --output=$(API_OUTPUT_DIR)
```

plugins/apidocs/apidocs.go
```diff
@@ -0,0 +1,212 @@
...
+type apidocsInstaller struct{}
+
+// Name is the external config name.
+func (i *apidocsInstaller) Name() string {
+	return ExternalConfigName
+}
+
+type config struct {
+	path  string
+	files string
```
kdada: Field `files string` should be `files map[string]string`.

plugins/apidocs/apidocs.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Name returns plugin name.
+func (p *Option) Name() string {
+	return ExternalConfigName
+}
+
+// Configure configures nirvana config via current options.
+func (p *Option) Configure(cfg *nirvana.Config) error {
+	cfg.Configure(
```
kdada: Then converts `Option.Files` to `config.files`.

plugins/apidocs/apidocs.go
```diff
@@ -0,0 +1,212 @@
...
+func (i *apidocsInstaller) Install(builder service.Builder, cfg *nirvana.Config) error {
+	var err error
+	wrapper(cfg, func(c *config) {
+		path, files := i.handle(c)
+
+		var routers []string
+		for k, v := range files {
+			routers = append(routers, k)
+
+			err = builder.AddDescriptor(definition.Descriptor{
```
kdada: Handle this error.

plugins/apidocs/apidocs.go
```diff
@@ -0,0 +1,212 @@
...
+						return ioutil.ReadFile(v)
+					},
+				}},
+			})
+		}
+
+		err = builder.AddDescriptor(definition.Descriptor{
+			Path:     path,
+			Consumes: []string{definition.MIMEAll},
+			Produces: []string{definition.MIMEAll},
```
kdada: Produce `json`.

plugins/apidocs/apidocs.go
```diff
@@ -0,0 +1,212 @@
...
+			Definitions: []definition.Definition{{
+				Method:  definition.Get,
+				Results: definition.DataErrorResults(""),
+				Function: func(ctx context.Context) ([]string, error) {
+					fmt.Println(routers)
```
kdada: Remove fmt.

plugins/apidocs/apidocs.go
```diff
@@ -0,0 +1,212 @@
...
+		path, files := i.handle(c)
+
+		var routers []string
+		for k, v := range files {
+			routers = append(routers, k)
+
+			err = builder.AddDescriptor(definition.Descriptor{
+				Path:     k,
+				Consumes: []string{definition.MIMEAll},
+				Produces: []string{definition.MIMEAll},
```
kdada: Produce `json`.

plugins/apidocs/apidocs.go
kdada: Use `Descriptor.Children` instead of `builder.AddDescriptor`.

bbbmj:
@kdada Fixed, PTAL

PR #237, opened 2018-09-05
chore(*): add flag dry-run

bbbmj:

Add your description

Fixes #

Reference to #

/cc @liubog2008 @kdada 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/nirvana/api/api.go
+++ b/cmd/nirvana/api/api.go
@@ -67,11 +67,13 @@ func newAPICommand() *cobra.Command {
 }
 
 type apiOptions struct {
+	DryRun bool
 	Serve  string
 	Output string
 }
 
 func (o *apiOptions) Install(flags *pflag.FlagSet) {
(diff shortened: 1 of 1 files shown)


ddysher:
what happens if we do not have this option?

bbbmj:
> what happens if we do not have this option?

Maybe the only thing we want to do is output swagger.json rather than serve ReDoc, 
another way is to use option `output` to prevent serve.

kdada:
If you pass `--serve="" --output=./`, it will skip serving and just generate `swagger.json`. If the default behavior is not appropriate, you should focus on `--serve` rather than add a new flag.

bbbmj:
> If you pass --serve="" --output=./, it will skip serving and just generate swagger.json. If the default behavior is not appropriate, you should focus on --serve rather than add a new flag.

Emmm....................LGTM

cc @ddysher @liubog2008 

/close

PR #263, opened 2019-02-01
bugfix: correct resp should not return an err

silva6:

correct resp should not return an err

**Release note**:

```release-note
NONE
```


kdada:
Could you write a test for this case?

kdada:
/lgtm 

I'll add tests later.

PR #280, opened 2019-04-23
fix(#278): prevent WriteHeader from automatically calling when hijack used

### LINKED ISSUE #278: Want upgrade http to websocket but always got warning WriteHeader

/kind support

**What happened**:

I am use `gorilla/websocket` to upgrade websocket, but always got warning:

```
2019/04/16 10:48:44 http: response.WriteHeader on hijacked connection from github.com/caicloud/nirvana/service.(*response).WriteHeader (context.go:199)
```

Provide a way to disable WriteHeader calling.

wutz:

It will not be calling `WriteHeader` when upgrade http to websocket. 

Nirvana always check `HeaderWritable` and call `WriteHeader` , this will cause the `net/http` output warning log.

Fixes #278 

/cc @kdada 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/service/context.go
+++ b/service/context.go
@@ -176,6 +176,7 @@ type response struct {
 	writer        http.ResponseWriter
 	statusCode    int
 	contentLength int
+	hijacked      bool
 }
 
 // For http.HTTPResponseWriter and HTTPResponseInfo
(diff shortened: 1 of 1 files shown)


kdada:
I think it's more graceful to add a hijacked field and check it in `HeaderWritable()`.

wutz:
@kdada

PR #283, opened 2019-08-09
refactor(*): fix typo and add comments and tweak

hezhizhen:

refactor(*): fix typo and add comments and tweak

Fixes #

Reference to #

/cc @bbbmj 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/nirvana.go
+++ b/nirvana.go
@@ -68,7 +68,7 @@ type Config struct {
 	configSet map[string]interface{}
 	// locked is for locking current config. If the field
 	// is not 0, any modification causes panic.
-	locked int32 // TODO: why not use bool?
+	locked int32
 }
 
 // lock locks config. If succeed, it will return ture.
--- a/nirvana.go
+++ b/nirvana.go
@@ -30,7 +30,7 @@ import (
 )
 
 // Server is a complete API server.
-// The server contains a router to handle all requests form clients.
+// The server contains a router to handle all requests from clients.
 type Server interface {
 	// Serve starts to listen and serve requests.
 	// The method won't return except an error occurs.
@@ -68,7 +68,7 @@ type Config struct {
 	configSet map[string]interface{}
 	// locked is for locking current config. If the field
 	// is not 0, any modification causes panic.
-	locked int32
+	locked int32 // TODO: why not use bool?
 }
 
 // lock locks config. If succeed, it will return ture.
@@ -129,7 +129,7 @@ func (c *Config) Set(name string, config interface{}) {
 	}
 }
 
-// forEach traverse all plugin configs.
+// forEach traverses all plugin configs.
 func (c *Config) forEach(f func(name string, config interface{}) error) error {
 	for name, cfg := range c.configSet {
 		if err := f(name, cfg); err != nil {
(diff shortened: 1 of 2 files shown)


nirvana.go
```diff
@@ -68,7 +68,7 @@ type Config struct {
 	configSet map[string]interface{}
 	// locked is for locking current config. If the field
 	// is not 0, any modification causes panic.
-	locked int32
+	locked int32 // TODO: why not use bool?
```
kdada: There is no CAS function for bool.

nirvana.go
```diff
@@ -169,13 +169,9 @@ func NewDefaultConfig() *Config {
...
-		ip:          "",
-		port:        8080,
-		logger:      &log.SilentLogger{},
-		filters:     []service.Filter{},
-		descriptors: []definition.Descriptor{},
-		modifiers:   []service.DefinitionModifier{},
-		configSet:   make(map[string]interface{}),
+		port:      8080,
+		logger:    &log.SilentLogger{},
+		configSet: make(map[string]interface{}),
```
kdada: Why do you remove these field initializers?

hezhizhen: Because the values assigned to them are default ones, which is not necessary to explicitly write.
Besides, nil slice can append items, while using nil maps will cause panic.

nirvana.go
hezhizhen: Ok.

nirvana.go
kdada: It's a convention and not a constraint. I recommend initialize everything in the constructor.

hezhizhen: It sounds reasonable, but what if there are too many fields, like 10 or more?

nirvana.go
kdada: Remove this comments then LGTM.

nirvana.go
kdada: It's not a usual case. We always have a lot of fields to initialize.

hezhizhen: For simplicity, I prefer to only write fields whose assigned values are not default ones.
Or is there a clear line between structs that all the fields should be explicitly initialized and those not?

kdada:
@ddysher Please fix the goveralls. It does not work.

nirvana.go
ddysher: do we agree upon on this change?

ddysher:
I have no clue what happened here...

```
$ goveralls -service=travis-ci -coverprofile $GO_COVER_PROFILE
Bad response status from coveralls: 405
<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>
The command "goveralls -service=travis-ci -coverprofile $GO_COVER_PROFILE" exited with 1.
```

hezhizhen:
@ddysher 我上一个 commit 的 CI 是通过了的，所以周五晚上或者周末有做什么改动吗？

bbbmj:
> 是不是上周刷过 GitHub 的一些配置

没啥相关性～

ddysher:
image

比较怀疑是不是这个服务挂了...

bbbmj:
重试下就成功了。。

ddysher:
昨晚试过... 估计抽风了
> 重试下就成功了。。

PR #286, opened 2019-08-19
feat(plugin): add a type parameter for the health check plugin

iawia002:

Add a type parameter for the health check plugin to support different check types, such as readiness requires more check steps than liveness.

If this change makes sense, I'll update the document soon

Fixes #

Reference to #

/cc @kdada 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/examples/getting-started/healthcheck/echo.go
+++ b/examples/getting-started/healthcheck/echo.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 
+	"github.com/caicloud/nirvana"
 	"github.com/caicloud/nirvana/config"
 	"github.com/caicloud/nirvana/definition"
 	"github.com/caicloud/nirvana/errors"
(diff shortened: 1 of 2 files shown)


plugins/healthcheck/healthz.go
```diff
@@ -29,9 +29,10 @@ func init() {
 }
 
 // HealthChecker checks if current server is healthy.
-type HealthChecker func(ctx context.Context) error
+// The `checkType` parameter indicates the type of health check, such as liveness or readiness.
+type HealthChecker func(ctx context.Context, checkType string) error
```
kdada: This change breaks the backward compatibility.

examples/getting-started/healthcheck/echo.go
```diff
@@ -0,0 +1,83 @@
...
+func Echo(ctx context.Context, msg string) (string, error) {
+	return msg, nil
+}
+
+func main() {
+	cmd := config.NewDefaultNirvanaCommand()
+	cmd.EnablePlugin(
+		healthcheck.NewOption(func(ctx context.Context, checkType string) error {
+			switch checkType {
+			case "liveness":
```
kdada: Use constant as instead.

plugins/healthcheck/healthz.go
```diff
@@ -60,8 +61,11 @@ func (i *healthcheckInstaller) Install(builder service.Builder, cfg *nirvana.Con
 			Consumes: []string{definition.MIMEAll},
 			Produces: []string{definition.MIMEAll},
 			Definitions: []definition.Definition{{
-				Method:   definition.Get,
-				Results:  []definition.Result{definition.ErrorResult()},
+				Method:  definition.Get,
+				Results: []definition.Result{definition.ErrorResult()},
+				Parameters: []definition.Parameter{
+					definition.QueryParameterFor("type", "the type of health check"),
```
kdada: Why not use path like `/healthz/liveness`

iawia002: Fully fixed path or regexp path?

I believe that not everyone needs different types of checks. If you don't need that, just ignore the type parameter and use `/healthz`. In some other cases, we may need more types of checks.

plugins/healthcheck/healthz.go
iawia002: Yes, I didn't find a better way to keep the backward compatibility. Any suggestion?

plugins/healthcheck/healthz.go
iawia002: ping, any response? We need this feature to distinguish between different types of health checks (in our case, `liveness` and `readiness`)

kdada: This is my point and not a problem. I think you need to solve the problem of `backward compatibility`.

iawia002: So how to support different type of health check without break anything?

kdada: Add a new Configurer maker `func CheckerWithType(checker HealthCheckerWithType) nirvana.Configurer`. Then make `func Checker(checker HealthChecker) nirvana.Configurer` as an alternate.

iawia002: I made some changes as you request, PTAL

kdada:
/lgtm

But it's a little complex. You can do it like this:

```
// Checker returns a configurer to set health checker.
func Checker(checker HealthChecker) nirvana.Configurer {
	if checker == nil {
		checker = defaultHealthChecker
	}
	return func(c *nirvana.Config) error {
		wrapper(c, func(c *config) {
			c.checker = func(ctx context.Context, checkType string) error {
                              checker(ctx)
                         }
		})
		return nil
	}
}
```

Issue #288, opened 2019-08-23
remove the deprecated `CloseNotify` function


iawia002:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind feature

**What happened**:

https://github.com/caicloud/nirvana/blob/854a4623c350ca2e71741cc93142f13e2b6afe8a/service/context.go#L208-L211

the original `CloseNotifier` interface was marked as `Deprecated` since Go 1.11, the user should use `http.Request.Context.Done()` as instead. Should we remove this too?

PS: this will trigger a lint problem (#287)

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

/cc @kdada

kdada:
Just do it.

PR #289, opened 2019-08-23
chore(*): migrate to golangci-lint

### LINKED ISSUE #287: migrate to golangci-lint
[golangci-lint](https://github.com/golangci/golangci-lint) is faster than [gometalinter](https://github.com/alecthomas/gometalinter), easier to use, and rich in configuration. And since `gometalinter` is DEPRECATED for now, we should migrate from `gometalinter` to `golangci-lint`

related: https://github.com/caicloud/nirvana/issues/31

/cc @ddysher @kdada

iawia002:

close #287 

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```


service/context.go
```diff
@@ -206,6 +206,8 @@ func (c *response) Flush() {
 }
 
 // CloseNotify is a disguise of http.response.CloseNotify().
+//
+// Deprecated: use `http.Request.Context.Done()` as instead.
 func (c *response) CloseNotify() <-chan bool {
```
kdada: The implementation is not correct. This method makes users have no way to check if the response supports http.CloseNotifier. Do you have time to fix this in a new PR?

kdada:
LGTM for the rest.

service/context.go
iawia002: So I deleted it directly in this PR? As mentioned in #288

kdada: I mean that the response may not implement `http.CloseNotifier`.

The method should be separated into two parts:
```
type response struct {...}


type responseWithCloseNotify struct {
    *response
}

func (c *responseWithCloseNotify) CloseNotify() <-chan bool {...}
```

Then wrap response with the new struct.

```

rw = &response{}

if _, ok:=rw.(http.CloseNotifier); ok {
    rw = &responseWithCloseNotify{rw}
}
```

If the method is removed from the go library in the future, the `response` still works.

iawia002: You mean still keep this function? As you replied `Just do it.` in #288 I thought we can just delete this function.

kdada: It seems that we need to retain it because of the behavior of go library.

iawia002: Fine, I will fix this in a new PR

PR #293, opened 2019-09-27
fix: nirvana api cmd write wrong container root on windows.

narrowizard:
nirvana panic when i run `nirvana api --serve=":8081"` on windows.  
```shell
# error log
INFO  0927-13:54:53.799+08 api.go:87 | No packages are specified, defaults to pkg/apis
# command-line-arguments
nirvana-generated573984343\main.go:20:37: non-hex character in escape sequence: s
nirvana-generated573984343\main.go:20:48: unknown escape sequence
FATAL 0927-13:55:06.890+08 api.go:61 | exit status 2
```
finally i find that APIBuilder.file func in `utils/builder` write wrong root path like `c:\some\directory` in generated `main.go`.  
this pr fixes the problem.  
but i think it only works in linux-based shell(like git-bash) on windows.

/cc @kdada 

**Release note**:

```release-note
fix `nirvana api` command fail on windows.  
```

### DIFF
--- a/utils/builder/builder.go
+++ b/utils/builder/builder.go
@@ -21,13 +21,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"go/types"
-	"html/template"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"strconv"
+	"text/template"
 
 	"github.com/caicloud/nirvana/utils/api"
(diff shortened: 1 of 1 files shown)


narrowizard:
image

generated main.go with invalid string `C:\Users\narro\Documents\code\go\src\gitlab.com\narrowizard\floral-admin-api`

kdada:
/lgtm

Thanks for your PR.

PR #307, opened 2020-04-24
feat(api): switch to swagger-ui

iawia002:

78764679-7a8a0e00-79b9-11ea-87e5-56addf10c425

Fixes #

Reference to #

/cc @supereagle @hezhizhen 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/nirvana/api/api.go
+++ b/cmd/nirvana/api/api.go
@@ -204,74 +204,41 @@ func (o *apiOptions) pathForVersion(version string) string {
 
 func (o *apiOptions) indexData(versions []string) ([]byte, error) {
 	index := `
-{{ $total := len .}}
-{{ $multipleVersions := gt $total 1 }}
+<!-- HTML for static distribution bundle build -->
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
-    <title>ReDoc Demo: Multiple apis</title>
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta charset="UTF-8">
+    <title>Swagger UI</title>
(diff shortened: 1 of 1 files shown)


cmd/nirvana/api/api.go
```diff
@@ -204,74 +204,41 @@ func (o *apiOptions) pathForVersion(version string) string {
 
 func (o *apiOptions) indexData(versions []string) ([]byte, error) {
 	index := `
-{{ $total := len .}}
-{{ $multipleVersions := gt $total 1 }}
+<!-- HTML for static distribution bundle build -->
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
hezhizhen, line 209: 万一有中文咋办？

cmd/nirvana/api/api.go
```diff
@@ -280,40 +247,36 @@ func (o *apiOptions) indexData(versions []string) ([]byte, error) {
...
 	tmpl, err := template.New("index.html").Parse(index)
 	if err != nil {
 		return nil, err
 	}
-	data := []struct {
+	data := make([]struct {
 		Name string
 		Path string
-	}{}
+	}, 0, len(versions))
```
hezhizhen, line 279:
```
data := make([]xxx, len(versions))
for i, v := range versions {
    data[i].Name = v
    data[i].Path = o.pathForVersion(v)
}
```

cmd/nirvana/api/api.go
iawia002: This actually doesn't mean you can't write Chinese in the HTML file, this attribute just tells the browser that the language of this page is English, and then the browser can do something based on this information, eg: auto-translate.

hezhizhen:
LGTM @supereagle

PR #309, opened 2020-04-26
refactor(metrics): standardize Prometheus metrics

lichuan0620:

This PR should be considered as a __breaking change__.

Refactored the metrics so that they meet the [K8 Instrumentation Guideline](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/instrumentation.md), the changes include:

1. Removed redundant labels.
2. Added a `component` label to all metrics; it takes the value of metric namespace.
3. Renamed according to the naming convention.
4. The Histogram type request duration metric use standard bucket sizes (which produce much more data, which will put more pressure on Prometheus, but is also more precise).
5. Changed the unit of the request duration metric from milliseconds to seconds.

As a result, the metrics should now look a lot like the APIServer request metrics. 

Also, removed the `Summary` type request latency metric: `Histogram` metrics function similarly to `Summary` metrics and are generally considered as the superior option, since it supports all quantile values instead of just 0.5, 0.9, and 0.99. This should balance out the extra data volume added by the refactored Histogram metric.

Fixes #

Reference to #

/cc @supereagle @ddysher @iawia002 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/plugins/metrics/metrics.go
+++ b/plugins/metrics/metrics.go
@@ -62,19 +62,9 @@ func newMetricsMiddleware(namespace string) definition.Middleware {
 		},
 		[]string{"method", "path"},
 	)
-	requestLatenciesSummary := prometheus.NewSummaryVec(
-		prometheus.SummaryOpts{
-			Namespace: namespace,
-			Name:      "request_latencies_summary",
-			Help:      "Response latency summary in milliseconds for each verb and resource.",
-			MaxAge:    time.Hour,
-		},
-		[]string{"method", "path"},
-	)
 
 	prometheus.MustRegister(requestCounter)
 	prometheus.MustRegister(requestLatencies)
-	prometheus.MustRegister(requestLatenciesSummary)
 
 	return func(ctx context.Context, next definition.Chain) error {
 		startTime := time.Now()
@@ -88,7 +78,6 @@ func newMetricsMiddleware(namespace string) definition.Middleware {
 
 		requestCounter.WithLabelValues(req.Method, path, getHTTPClient(req), req.Header.Get("Content-Type"), strconv.Itoa(resp.StatusCode())).Inc()
 		requestLatencies.WithLabelValues(req.Method, path).Observe(elapsed)
-		requestLatenciesSummary.WithLabelValues(req.Method, path).Observe(elapsed)
 
 		return err
 	}
--- a/plugins/metrics/metrics.go
+++ b/plugins/metrics/metrics.go
@@ -18,7 +18,6 @@ package metrics
 
 import (
 	"context"
-	"net/http"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lichuan0620:
I'm submitting this PR just so that we can start discussing it. It does not mean that we should merge it immediately. It is a breaking change and it causes Nirvana to put more pressure on Prometheus under most circumstances (to make the Histogram metrics more precise and meaningful).

plugins/metrics/metrics.go
```diff
@@ -45,36 +44,29 @@ type config struct {
...
 		prometheus.CounterOpts{
-			Namespace: namespace,
-			Name:      "request_count",
-			Help:      "Counter of server requests broken out for each verb, API resource, client, and HTTP response contentType and code.",
+			Namespace:   namespace,
+			Name:        "request_total",
+			Help:        "Counter of server requests broken out for each verb, API resource, and HTTP response code.",
+			ConstLabels: constLabel,
 		},
-		[]string{"method", "path", "client", "contentType", "code"},
```
ddysher, line 54: so `client` and `contentType` is not needed any more?

lichuan0620: I don't think they are important enough to justify the additional dimension. Also, I've never seen any open-sourced project adding those to their request related metrics.

lichuan0620:
> What are the benefits of meeting Kubernetes guideline for Nirvana, apart from consistency?

Consistency by itself is a major benefit. In any production setup, one would face hundreds if not thousands of metrics. There is no way you would read the description on each and every one of them. By following the convention, you can understand the metrics by looking at their names.

For example, the first word of a metric name is usually its targeting component, and if a metric end with a `_total` postfix, then it's a counter.

If also allow you to do something like  `{__name__=~"(.+)request_duration_(.+)"}` to pull out request latency metrics from all components.

Non-naming conventions, on the other hand, are usually good for performance or/and precision. For example, the removed redundant labels save memory and make queries faster. The added Histogram buckets make `histogram_quantile` function much more predictable and precise.

> Are we actively using those Summary-based metrics now?

We are not using any Nirvana metrics right now. Even if we are, Summary is totally replaceable by Histogram so it doesn't matter. As for Kubernetes, almost all Summary-typed metrics have been depreciated as of `v1.17.0`. Other open-sourced projects are slower and more unwilling to adapt.

iawia002:
The code LGTM, but since I'm not familiar with Prometheus, I can't give more comments about this. And if there is no more comment, I think we can merge it into the next release (0.3.x).

plugins/metrics/metrics.go
ddysher: SGTM

ddysher:
LGTM

@lichuan0620 thanks for the detailed explanation!

supereagle:
LGTM

ddysher:
@iawia002 make sure to make it explicit in the release note.

> This PR should be considered as a breaking change.

iawia002:
> @iawia002 make sure to make it explicit in the release note.
> 
> > This PR should be considered as a breaking change.

Will do

PR #313, opened 2020-04-28
feat(middleware): implemented two reqlog middlewares

lichuan0620:

Added two middleware that mimick the functionality of the `reqlog` plugin. Unlike a plugin, with these middlewares, the user would be able to enable request log only for a subset of their descriptors.

Fixes #

Reference to #

/cc @iawia002 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/middlewares/reqlog/reqlog.go
+++ b/middlewares/reqlog/reqlog.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2017 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reqlog
+
+import (
+	"context"
+	"time"
+
+	"github.com/caicloud/nirvana/definition"
+	"github.com/caicloud/nirvana/log"
+	"github.com/caicloud/nirvana/service"
+)
+
+// Default returns a reqlog middleware that uses the default Nirvana logger.
+func Default() definition.Middleware {
+	return func(ctx context.Context, chain definition.Chain) error {
+		start := time.Now()
+		httpCtx := service.HTTPContextFrom(ctx)
+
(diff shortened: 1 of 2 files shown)


lichuan0620:
Moved from go-common.

I plan to also implement a metric middleware after #309 is merged.

middlewares/reqlog/reqlog.go
```diff
@@ -0,0 +1,70 @@
...
+func Default() definition.Middleware {
+	return func(ctx context.Context, chain definition.Chain) error {
+		start := time.Now()
+		httpCtx := service.HTTPContextFrom(ctx)
+
+		err := chain.Continue(ctx)
+
+		request := httpCtx.Request()
+		response := httpCtx.ResponseWriter()
+		log.Infoln(
```
supereagle, line 38: Different type of logs according to the response status code? It will be helpful when we want to analyse these logs.

middlewares/reqlog/reqlog.go
```diff
@@ -0,0 +1,70 @@
...
+func Custom(logger log.Logger, level log.Level) definition.Middleware {
+	return func(ctx context.Context, chain definition.Chain) error {
+		start := time.Now()
+		httpCtx := service.HTTPContextFrom(ctx)
+
+		err := chain.Continue(ctx)
+
+		request := httpCtx.Request()
+		response := httpCtx.ResponseWriter()
+		logger.V(level).Infoln(
```
supereagle, line 60: ditto

middlewares/reqlog/reqlog.go
lichuan0620: Then this middleware would behave differently than the `reqlog` plugin. Maybe we can discuss later? If we wanted the change, we should change it everywhere.

supereagle: Do we need to change any others besides `reqlog` plugin? If not, I think we can change it.
/cc @iawia002

lichuan0620: > Do we need to change any others besides reqlog plugin

No, `reqlog` is the only other place where we are doing this. Still, I think this could use some discussion, perhaps a proposal to start it. There are several drawbacks I could think of:

1. Should we treat 4xx responses as errors? The service might not care about client-side errors.
2. People who are not building a restful service might find it annoying. They might want to redefine the meaning of their status codes.

After all, Nirvana is a framework. I'm not sure if it's wise to be too assertive.

iawia002: I'm not so sure if we need different log types. For users who want more detailed log, I bet they will not even use this middleware, they will manually log where needed with custom content.

Issue #297, opened 2020-02-19
generate API doc failed: goroutine stack exceeds 1000000000-byte limit


tskdsb:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug

**What happened**:
```
nirvana api /root/tsk/src/github.com/caicloud/clever-admin/pkg/v2/apis >log
WARN  0219-15:52:25.066+08 builder.go:69 | can't find project file, instead by default config


runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0x1bf9c99, 0xe)
	/root/tsk/go/src/runtime/panic.go:774 +0x72
runtime.newstack()
	/root/tsk/go/src/runtime/stack.go:1046 +0x6e9
runtime.morestack()
	/root/tsk/go/src/runtime/asm_amd64.s:449 +0x8f

goroutine 1 [running]:
reflect.(*rtype).Name(0x1871a60, 0x0, 0x0)
	/root/tsk/go/src/reflect/type.go:874 +0xa8 fp=0xc020930308 sp=0xc020930300 pc=0x483098
github.com/caicloud/nirvana/utils/api.(*TypeContainer).NameOf(0xc000845da0, 0x1efcba0, 0x1871a60, 0x4493ff, 0x30)
	/root/tsk/src/github.com/caicloud/clever-admin/vendor/github.com/caicloud/nirvana/utils/api/types.go:163 +0x4f fp=0xc0209304f0 sp=0xc020930308 pc=0x16db1af
github.com/caicloud/nirvana/utils/api.(*TypeContainer).NameOf(0xc000845da0, 0x1efcba0, 0x18d86a0, 0x3, 0x0)
	/root/tsk/src/github.com/caicloud/clever-admin/vendor/github.com/caicloud/nirvana/utils/api/types.go:187 +0x776 fp=0xc0209306d8 sp=0xc0209304f0 pc=0x16db8d6
github.com/caicloud/nirvana/utils/api.(*TypeContainer).NameOf(0xc000845da0, 0x1efcba0, 0x1a10e40, 0x16ea14c, 0x6)
	/root/tsk/src/github.com/caicloud/clever-admin/vendor/github.com/caicloud/nirvana/utils/api/types.go:207 +0x281 fp=0xc0209308c0 sp=0xc0209306d8 pc=0x16db3e1
[...]

PR #319, opened 2020-05-08
fix(utils): handle the nested element that may cause infinite recursion when generating API docs

iawia002:

Handle struct fields separately to deal with nested structures which may cause infinite recursion when generating API docs, eg: 

```go
type Message struct {
	ID      int                  `json:"id"`
	Title   []string             `json:"title"`
	Content [][]string           `json:"content"`
	M1      *[]*[]*Message       `json:"m1"`
	M2      *Message             `json:"m2"`
	M3      map[*Message]Message `json:"m3"`
}
```

This will cause infinite recursion when nirvana trying to fetch the info of the field `M1`, `M2` and `M3`, after this change, the `Message` struct in `M1`, `M2` and `M3` will be treated as empty:

Screen Shot 2020-05-08 at 15 02 48

Reference to #297 

/cc @supereagle @hezhizhen @qianlei90 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/utils/api/types.go
+++ b/utils/api/types.go
@@ -144,9 +144,9 @@ func (tc *TypeContainer) setType(t *Type, typ reflect.Type) {
 	}
 	tc.lock.Lock()
 	defer tc.lock.Unlock()
-	types := tc.real[tn]
-	index := len(types)
-	for i, originalType := range types {
+	typs := tc.real[tn]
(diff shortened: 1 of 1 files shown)


iawia002:
ping @supereagle @hezhizhen

hezhizhen:
LGTM

PR #321, opened 2020-05-11
chore(*): downgrade swagger-ui version to 3.22.2

iawia002:

After we support recursive definition in #319, I found some special structures will cause swagger-ui to report an error (data presentation is normal, just the web page keeps prompting for this error), eg:

```go
type T2 struct {
	ID  int `json:"id"`
	T21 *T2 `json:"t21"`
}

// Message describes a message entry.
type Message struct {
	ID      int                  `json:"id"`
	Title   []string             `json:"title"`
	Content [][]string           `json:"content"`
	T2      T2                   `json:"t2"`
}
```

Screen Shot 2020-05-08 at 18 20 04

After a period of time of debugging, I found out that this issue may be caused by swagger-client@3.9.0 (introduced in version 3.22.3), so this PR downgrades swagger-ui version to 3.22.2 and everything works correctly again.

I'll continue to look further to see if a new version addresses this issue or if we need to change something. Some related issues:

https://github.com/swagger-api/swagger-ui/issues/4214
https://github.com/swagger-api/swagger-js/issues/1362
https://github.com/swagger-api/swagger-ui/issues/4787

Fixes #

Reference to #

/cc @hezhizhen @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/nirvana/api/api.go
+++ b/cmd/nirvana/api/api.go
@@ -210,9 +210,9 @@ func (o *apiOptions) indexData(versions []string) ([]byte, error) {
   <head>
     <meta charset="UTF-8">
     <title>Swagger UI</title>
-    <link rel="stylesheet" type="text/css" href="https://unpkg.com/swagger-ui-dist@3.25.0/swagger-ui.css" >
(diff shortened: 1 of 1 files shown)


hezhizhen:
LGTM

Issue #272, opened 2019-03-21
Auto parameter string default value without double quotation make api docs generating failed


wutzx:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

`nirvana api` starting failed with logging:

FATAL 0321-12:50:40.034+08 api.go:61 | json: error calling MarshalJSON for type spec.Swagger: json: error calling MarshalJSON for type *spec.Paths: json: error calling MarshalJSON for type spec.PathItem: json: error calling MarshalJSON for type *spec.Operation: json: error calling MarshalJSON for type spec.OperationProps: json: error calling MarshalJSON for type spec.Parameter: json: error calling MarshalJSON for type *swagger.rawJSON: invalid character 'w' looking for beginning of value

**What you expected to happen**:

`nirvana api` can work correctly.

**How to reproduce it (as minimally and precisely as possible)**:

Define a struct which relating to Auto parameter, set string default value without double quotation:

```go
type Option struct {
  Hello string `source:"Query,hello,default=world"`
}
```

**Anything else we need to know?**:

If set default value with double quotation, then `nirvana api` can work, but bussiness function will be getting addition double quotation from that option.

```go
type Option struct {
  Hello string `source:"Query,hello,default=\"world\""`
}
```

kdada:
For non-basic types, you can register custom converters by `service.RegisterConverter()`.

wutzx:
But it does not work with api docs generating by resgister converter.

Wei Guo <notifications@github.com>于2019年3月21日 周四下午7:09写道：

wutzx:
The marjor issue is normal parameter default value coming from json
unencoding which will including double quotation for string type or base
string type, but for auto parameter default value coming from tag default
raw value.
Taizeng Wu <wutaizeng@gmail.com>于2019年3月21日 周四下午7:13写道：

kdada:
It's a problem. I have not found an effective way to make them work together.

PR #333, opened 2020-06-01
fix(api-docs): handle the default value of parameter correctly

iawia002:

For now, if we set a default value for a field of an auto parameter, the API docs generation will fail with the error below:

```
FATAL 0601-16:59:25.166+08 api.go:61 | json: error calling MarshalJSON for type spec.Swagger: json: error calling MarshalJSON for type *spec.Paths: json: error calling MarshalJSON for type spec.PathItem: json: error calling MarshalJSON for type *spec.Operation: json: error calling MarshalJSON for type spec.OperationProps: json: error calling MarshalJSON for type spec.Parameter: json: error calling MarshalJSON for type *swagger.rawJSON: invalid character 's' after top-level value
```

Fixes #

Reference to #272 

/cc @hezhizhen @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/utils/generators/swagger/generator.go
+++ b/utils/generators/swagger/generator.go
@@ -17,6 +17,8 @@ limitations under the License.
 package swagger
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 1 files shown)


utils/generators/swagger/generator.go
```diff
@@ -495,14 +534,22 @@ func (g *Generator) enum(typ *api.Type) []spec.Parameter {
 		parameters := []spec.Parameter(nil)
 		if tag != "" {
 			source, name, apc, err := service.ParseAutoParameterTag(tag)
-			defaultValue := apc.Get(service.AutoParameterConfigKeyDefaultValue)
+			rawDefaultValue := apc.Get(service.AutoParameterConfigKeyDefaultValue)
+			var defaultValue []byte
+			if c := converterFor(string(field.Type)); rawDefaultValue != "" && c != nil {
+				// we didn't find a good way to handle the default value of none-basic types,
```
hezhizhen, line 540: didn't? don't?

utils/generators/swagger/generator.go
```diff
@@ -485,7 +487,44 @@ func (g *Generator) generateAutoParameter(typ api.TypeName) []spec.Parameter {
 		return nil
 	}
 	return g.enum(structType)
+}
 
+func converterFor(typ string) service.Converter {
```
hezhizhen, line 492: 这个直接是 map 变量和一个函数有啥区别吗？

utils/generators/swagger/generator.go
```diff
@@ -495,14 +534,22 @@ func (g *Generator) enum(typ *api.Type) []spec.Parameter {
 		parameters := []spec.Parameter(nil)
 		if tag != "" {
 			source, name, apc, err := service.ParseAutoParameterTag(tag)
-			defaultValue := apc.Get(service.AutoParameterConfigKeyDefaultValue)
+			rawDefaultValue := apc.Get(service.AutoParameterConfigKeyDefaultValue)
+			var defaultValue []byte
+			if c := converterFor(string(field.Type)); rawDefaultValue != "" && c != nil {
```
hezhizhen, line 539: raw != nil && c == nil, default 就是空的吗？

utils/generators/swagger/generator.go
iawia002: 我再改一下，刚开始这个函数还有一些其它的东西

utils/generators/swagger/generator.go
iawia002: 对，空默认值，不直接报错

utils/generators/swagger/generator.go
iawia002: Updated

iawia002:
All comments addressed, PTAL @hezhizhen

hezhizhen:
/lgtm
but I'm not sure if it's expected...

iawia002:
> but I'm not sure if it's expected...

What? Setting an empty default value for those non-basic types?

utils/generators/swagger/generator.go
```diff
@@ -495,14 +531,22 @@ func (g *Generator) enum(typ *api.Type) []spec.Parameter {
 		parameters := []spec.Parameter(nil)
```
hezhizhen, line 531: 这样子写的话，一开始长度就是 1？

iawia002: 0

hezhizhen: 噢，是小括号，不是大括号。

iawia002:
If default values of non-basic types are not handled, the effect is as follows, the Swagger UI will not display the default values, that's all, I think it's better than returning errors directly during document generation.

Screen Shot 2020-06-02 at 12 06 44

So any more comments?

hezhizhen:
> If default values of non-basic types are not handled, the effect is as follows, the Swagger UI will not display the default values, that's all, I think it's better than returning errors directly during document generation.
> 
> Screen Shot 2020-06-02 at 12 06 44
> 
> So any more comments?

works for me.

PR #341, opened 2020-07-03
fix(template): enable race detection in tests only

iawia002:

Do not enable the `-race` flag in production, it will affect performance.

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```


cmd/nirvana/project/init.go
```diff
@@ -836,7 +836,7 @@ $(GOLANGCI_LINT):
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(BIN_DIR) v1.23.6
 
 test:
-	@go test -coverprofile=coverage.out ./...
+	@go test -race -coverprofile=coverage.out ./...
```
hezhizhen, line 839: 把上面 race 的注释移到这一行的上面

iawia002: Nice catch

hezhizhen:
LGTM

PR #342, opened 2020-07-14
feat(*): support to add a http.Handler directly to the router

### LINKED ISSUE #339: [question] Can we add a http.Handler (e.g. httputil.ReverseProxy) directly to the router?

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind feature

**What happened**:

Hi, thanks for the awesome RESTful framework! It helps me a lot. 

I have some problems when I tried to add a reverse proxy to the nirvana router. What I want to do is like the pseudo code here:

```go
	root.NewRoute().
		Method(http.MethodPut).
		Path("/manifests/:reference").
		HandlerFunc(putManifest)
	// initiate blob upload
	root.NewRoute().
		Method(http.MethodPost).
		Path("/blobs/uploads").
		Handler(reversePoxy)
```

We want to forward some requests to the `httputil.ReverseProxy` (A struct implements http.Handler interface) according to the path. I looked through the code base of nirvana and not sure if we can do it.

We compile the router tree like this now:

```go
func (b *builder) Build() (Service, error) {
	if len(b.bindings) <= 0 {
		return nil, noRouter.Error()
	}
	var root router.Router
	for path, bd := range b.bindings {
		b.logger.V(log.LevelDebug).Infof("Definitions: %d Middlewares: %d Path: %s",
			len(bd.definitions), len(bd.middlewares), path)
		top, leaf, err := router.Parse(path)
		if err != nil {
			b.logger.Errorf("Can't parse path: %s, %s", path, err.Error())
			return nil, err
		}
		if len(bd.definitions) > 0 {
			// RedirectTrailingSlash would redirect "/somepath/" to "/somepath". Any definition under "/somepath/"
			// will never be executed.
			if len(path) > 1 && strings.HasSuffix(path, "/") {
				b.logger.Warningf("If RedirectTrailingSlash filter is enabled, following %d definition(s) would not be executed", len(bd.definitions))
			}
			inspector := newInspector(path, b.logger)
			for _, d := range bd.definitions {
				b.logger.V(log.LevelDebug).Infof("  Method: %s Consumes: %v Produces: %v",
					d.Method, d.Consumes, d.Produces)
				if b.modifier != nil {
					b.modifier(&d)
				}
				if err := inspector.addDefinition(d); err != nil {
					return nil, err
				}
			}

			leaf.SetInspector(inspector)
		}
		for _, m := range bd.middlewares {
			m := m
			leaf.AddMiddleware(func(ctx context.Context, chain router.RoutingChain) error {
				return m(ctx, chain)
			})
		}
		if root == nil {
			root = top
		} else if root, err = root.Merge(top); err != nil {
			return nil, err
		}
	}
	s := &service{
		root:      root,
		filters:   b.filters,
		logger:    b.logger,
		producers: AllProducers(),
	}
	return s, nil
}
```

The service in the code implements http.Handler interface:

```go
func (s *service) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
	for _, f := range s.filters {
		if !f(resp, req) {
			return
		}
	}
	ctx := newHTTPContext(resp, req)

	executor, err := s.root.Match(ctx, &ctx.container, req.URL.EscapedPath())
	if err != nil {
		if err := writeError(ctx, s.producers, err); err != nil {
			s.logger.Error(err)
		}
		return
	}
	err = executor.Execute(ctx)
	if err == nil && ctx.response.HeaderWritable() {
		err = invalidService.Error()
	}
	if err != nil {
		if ctx.response.HeaderWritable() {
			if err := writeError(ctx, s.producers, err); err != nil {
				s.logger.Error(err)
			}
		} else {
			s.logger.Error(err)
		}
	}
}
```

Seems that we do not support such use case. I am not sure if I understand the logic right, thus glad to open an issue here to see if it is true.

I'd appreciate it if you could help.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

iawia002:

This PR makes nirvana supports to add a http.Handler (e.g. httputil.ReverseProxy) directly to the router.

See examples/getting-started/handler for a real example.

close #339 

/cc @gaocegege 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/definition/definition.go
+++ b/definition/definition.go
@@ -18,6 +18,7 @@ package definition
 
 import (
 	"context"
+	"net/http"
 	"reflect"
 )
 
@@ -202,6 +203,11 @@ type Definition struct {
 	// In some cases, succeessful data and error data should be generated in
 	// different ways.
 	ErrorProduces []string
+	// Handler is a http.Handler implementation, nirvana supports to add a http.Handler (e.g. httputil.ReverseProxy)
+	// directly to the router.
+	// If this field is not nil, there is no need to set the Function, Parameters, and Results fields.
+	// See examples/getting-started/handler for a real example.
+	Handler http.Handler
 	// Function is a function handler. It must be func type.
 	Function interface{}
 	// Parameters describes function parameters.
--- a/examples/getting-started/handler/main.go
+++ b/examples/getting-started/handler/main.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2020 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/http/httputil"
+	"net/url"
+
+	"github.com/caicloud/nirvana/config"
+	"github.com/caicloud/nirvana/definition"
+	"github.com/caicloud/nirvana/log"
+	"github.com/caicloud/nirvana/middlewares/reqlog"
+)
+
+func main() {
+	backendServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "this call was relayed by the reverse proxy")
+	}))
+	defer backendServer.Close()
+
+	rpURL, err := url.Parse(backendServer.URL)
+	if err != nil {
+		log.Fatal(err)
+	}
+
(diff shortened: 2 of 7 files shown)


gaocegege:
Is the feature compatible with middlewares?

iawia002:
> Is the feature compatible with middlewares?

Yes, I'll update the example and add a reqlog middleware for it, running the example and you'll see:

```
INFO  0714-16:11:48.691+08 config.go:309 | Listening on :8080
INFO  0714-16:11:48.691+08 builder.go:241 | Definitions: 1 Middlewares: 0 Path: /
INFO  0714-16:11:48.691+08 builder.go:256 |   Method: Get Consumes: [*/*] Produces: [application/json]
INFO  0714-16:11:48.691+08 builder.go:241 | Definitions: 1 Middlewares: 1 Path: /proxy
INFO  0714-16:11:48.691+08 builder.go:256 |   Method: Get Consumes: [*/*] Produces: [*/*]
INFO  0714-16:12:44.687+08 reqlog.go:38 | GET 200 43 24.944034ms /proxy
INFO  0714-16:12:53.720+08 reqlog.go:38 | GET 200 43 4.874487ms /proxy
```

iawia002:
Since there are no other comments, I'm gonna merge this. Please try this with the master branch, fire an issue if there are any bugs.

PR #347, opened 2020-07-17
feat(*): add a Optional field to the Parameter struct

### LINKED ISSUE #345: Guarantee not return error when use File type parameter.

> Uncomment only one, leave it on its own line:
>

/kind feature

**What happened**:

When use [File](https://github.com/caicloud/config-admin/blob/d8dae3eda049834e683de4bb1bc69131a3a02900/pkg/api/v1/descriptor/config.go#L90) type Parameters, it will return error `required field file in File but got empty` to client when client don't set the parameter, but the file parameter is optional.

Guarantee dont't return error even if the parameter is empty.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

iawia002:

fix #345

After this, you can set the `Optional` field of the parameter to true and handle the empty parameter situation yourself, eg:

```go
Parameters: []definition.Parameter{
	{
		Source:   definition.File,
		Name:     "file",
		Optional: true,
	},
},

func Upload(ctx context.Context, file multipart.File) (string, error) {
	if file == nil {
		// do something
	}
	// do something else
}
```

Fixes #

Reference to #

/cc @whalecold 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/service/executor.go
+++ b/service/executor.go
@@ -472,9 +472,8 @@ func (e *executor) Execute(ctx context.Context) (err error) {
 				return writeError(ctx, e.errorProducers, err)
 			}
 		}
-		if result == nil {
-			return writeError(ctx, e.errorProducers, requiredField.Error(p.name, p.generator.Source()))
-		} else if closer, ok := result.(io.Closer); ok {
+
+		if closer, ok := result.(io.Closer); ok {
 			defer func() {
 				if e := closer.Close(); e != nil && err == nil {
 					// Need to print error here.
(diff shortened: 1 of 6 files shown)


whalecold:
/lgtm
@hezhizhen

whalecold:
Thanks !

hezhizhen:
Here is another solution: add a boolean field `Required` to parameter which indicates if the field must be provided. 

Suppose that one parameter is required and a user forgets to provide it when calling the API, then nirvana will throw an error. If the validation is removed, we need to add the process to every API manually, which is quite redundant.

service/executor.go
```diff
@@ -483,7 +488,11 @@ func (e *executor) Execute(ctx context.Context) (err error) {
 			}()
 		}
 
-		paramValues = append(paramValues, reflect.ValueOf(result))
+		if result == nil {
+			paramValues = append(paramValues, reflect.New(p.targetType).Elem())
```
hezhizhen, line 492: 这里你试过 targetType 分别为 multipart.File 和 *multipart.File 吗？都没有问题？

iawia002: `multipart.File` 是个 interface，如果它是 struct 的话，不是指针的话它不会是 nil，会是一个空的 struct

PR #373, opened 2020-10-27
feat(cmd): add version command

knight42:

Add command `nirvana version`, so that users are able to know which version of nirvana they are using.

Fixes #

Reference to #

/cc @iawia002 @gaocegege 

**Release note**:

```release-note
cmd: add command `nirvana version`
```


iawia002:
This is probably not working, we normally install nirvana by `go get` command and we never publish any binary file. Maybe we can hardcode the version into the help message.

knight42:
@iawia002 I think we could probably build the binary on travis and upload the asset to GitHub automatically.

iawia002:
However, it is easier and more convenient for users to install with `go get`, so perhaps hardcode the version information would be a more general solution.

xpofei:
This way of printing version is ok. However you still should consider the output of 'nirvana version' when using manual 'go build'. In this case, it is recommended to print prompt statements such as "please use make tool to build in order to use version conmmand" instead of print a raw "v0.0.0-master+$Format:%h$"

Makefile
```diff
@@ -35,3 +42,12 @@ refine: format license
...
 	@gitbook build ./manuals ./docs
+
+.PHONY: build-local
+build-local:
+	CGO_ENABLED=0 go build -trimpath $(GOFLAGS) -o bin/nirvana -ldflags "-s -w \
+	-X $(VERSION_PKG).version=$(VERSION) \
+	-X $(VERSION_PKG).branch=$(BRANCH) \
+	-X $(VERSION_PKG).gitCommit=$(GITCOMMIT) \
+	-X $(VERSION_PKG).gitTreeState=$(GITTREESTATE) \
+	-X $(VERSION_PKG).buildDate=$(BUILDDATE)" ./cmd/nirvana
```
gaocegege, line 53: Should we also change the go template? cc @iawia002

cmd/nirvana/version/version.go
```diff
@@ -0,0 +1,80 @@
...
+	version      = "v0.0.0-master+$Format:%h$"
+	branch       = ""
+	gitCommit    = "$Format:%H$"          // sha1 from git, output of $(git rev-parse HEAD)
+	gitTreeState = ""                     // state of git tree, either "clean" or "dirty"
+	buildDate    = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
+)
+
+// Info contains versioning information.
+type Info struct {
+	Version      string `json:"version"`
```
gaocegege, line 51: Should we add omitempty?

Makefile
iawia002: > go template

Can you be more specific?

knight42:
/close
in favor of #376

PR #376, opened 2020-10-28
feat(cmd): add version information

iawia002:

See previous discussions in https://github.com/caicloud/nirvana/pull/373#issuecomment-717156181, this is another simpler solution, users are able to get the version information no matter how they install nirvana.

```console
$ nirvana
Nirvana toolchains

Usage:
  nirvana [command]

Available Commands:
  api         Generate API documents for your project
  client      Create client for project
  help        Help about any command
  init        Create a basic project structure

Flags:
  -h, --help      help for nirvana
      --version   version for nirvana

Use "nirvana [command] --help" for more information about a command.

$ nirvana --version
nirvana version v0.3.0-alpha.0
```

I'll manually update the version number when we are going to release a new version.

Fixes #

Reference to #

/cc @knight42 @xpofei 

**Release note**:

```release-note
cmd: add version information.
```

### DIFF
--- a/cmd/nirvana/main.go
+++ b/cmd/nirvana/main.go
@@ -21,12 +21,14 @@ import (
 	"github.com/caicloud/nirvana/cmd/nirvana/client"
 	"github.com/caicloud/nirvana/cmd/nirvana/project"
 	"github.com/caicloud/nirvana/log"
+
 	"github.com/spf13/cobra"
 )
 
 var root = &cobra.Command{
-	Use:   "nirvana",
-	Short: "Nirvana toolchains",
+	Use:     "nirvana",
+	Short:   "Nirvana toolchains",
+	Version: "v0.3.0",
 }
 
 func main() {


cmd/nirvana/main.go
```diff
@@ -21,12 +21,14 @@ import (
...
+
 	"github.com/spf13/cobra"
 )
 
 var root = &cobra.Command{
-	Use:   "nirvana",
-	Short: "Nirvana toolchains",
+	Use:     "nirvana",
+	Short:   "Nirvana toolchains",
+	Version: "v0.3.0-alpha.0",
```
knight42, line 31: IMHO I prefer automation than manual operation, the one who is responsible for making a release might forget to update the info here before he/she creates a new release.

In addition, `kubectl`, `kind`, `helm`(including but not limited to) these commands all provide a `version` subcommand, and they all inject the version information at build time:
* kubectl:  https://github.com/kubernetes/kubernetes/blob/8456a514f18925a4406e385f32b0a9fc659b7966/hack/lib/version.sh#L165-L178
* kind: https://github.com/kubernetes-sigs/kind/blob/86bcbff17fb2afff93f56479cccb97812970e67a/Makefile#L55
* helm: https://github.com/helm/helm/blob/82f739072cb3c94f26e3103e4524a1f74526342d/Makefile#L53-L56

I think this is a good convention to follow.

iawia002: But nirvana is just a simple and small project, you can simply install it by `go get` rather than downloading a binary file or executing another script, for other users who install it with `go get`, they can't even get the version information.

Similar tooling projects like Hugo are also hardcode the version information: https://github.com/gohugoio/hugo/blob/master/common/hugo/version_current.go

PR #382, opened 2020-11-19
feat(*): support RPC style

iawia002:
**What this PR does / why we need it**:

Add a new builder to support serve RPC style APIs:

```
POST /?Action=GetUserByID&Version=2018-01-01
```

The default builder is still RESTful builder, here is an example to use RPC builder:

```go
var echo = definition.RPCDescriptor{
	Description: "Echo API",
	Middlewares: []definition.Middleware{reqlog.Default()},
	Consumes:    []string{definition.MIMEAll},
	Produces:    []string{definition.MIMEJSON},
	Actions: []definition.RPCAction{
		{
			Name:    "Echo",
			Version: "2020-10-10",
			Function: func(ctx context.Context, msg string) (string, error) {
				return msg, nil
			},
			Parameters: []definition.Parameter{
				{
					Source:      definition.Query,
					Name:        "msg",
					Description: "Corresponding to the second parameter",
				},
			},
			Results: definition.DataErrorResults(""),
		},
		{
			Name:    "Echo1",
			Version: "2020-10-10",
			Function: func(ctx context.Context, msg string) (string, error) {
				return msg, nil
			},
			Parameters: []definition.Parameter{
				{
					Source:      definition.Query,
					Name:        "msg",
					Description: "Corresponding to the second parameter",
				},
			},
			Results: definition.DataErrorResults(""),
		},
	},
}

func main() {
	cmd := config.NewDefaultNirvanaCommand()
	conf := nirvana.NewDefaultConfig()
	conf.Configure(
		nirvana.APIStyle(builder.APIStyleRPC),
		nirvana.Descriptor(echo, echo2),
	)
	if err := cmd.ExecuteWithConfig(conf); err != nil {
		log.Fatal(err)
	}
}
```

```
INFO  1119-18:25:59.051+08 config.go:309 | Listening on :8080
INFO  1119-18:25:59.052+08 builder.go:180 | Path: /?Version=2020-10-10&Action=Echo, Consumes: [*/*], Produces: [application/json]
INFO  1119-18:25:59.052+08 builder.go:180 | Path: /?Version=2020-10-10&Action=Echo1, Consumes: [*/*], Produces: [application/json]
```

And for generating API documentation for RPC style APIs, you need to add a `style` tag to the descriptors file:

```
// +nirvana:api=descriptors:"Descriptor" style:"rpc"
```

Screen Shot 2020-11-19 at 18 41 33

**Which issue(s) this PR is related to** *(optional, link to 3rd issue(s))*:

Fixes #

[...]

PR #390, opened 2020-11-22
feat(*): make multiple plugins support the RPC builder

iawia002:

Other plugins like metrics rely on configuring the middleware for the root path (/) to automatically configure the middleware for all paths which doesn't work for the RPC API style, which requires manual configuration of middleware and descriptor for the RPC style server.

Fixes #

Reference to #382 

/cc @xpofei 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/service/rpc/builder.go
+++ b/service/rpc/builder.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"sort"
 
 	"github.com/caicloud/nirvana/definition"
 	"github.com/caicloud/nirvana/log"
@@ -88,6 +89,9 @@ func (b *builder) SetModifier(m service.DefinitionModifier) {
 }
 
 func genRPCPath(path, version, action string) string {
+	if version == "" && action == "" {
+		return path
+	}
 	return fmt.Sprintf("%s?Version=%s&Action=%s", path, version, action)
 }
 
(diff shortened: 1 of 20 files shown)


utils/api/definitions.go
```diff
@@ -100,7 +99,7 @@ type Definition struct {
 // NewDefinition creates openapi.Definition from definition.Definition.
 func NewDefinition(tc *TypeContainer, d *definition.Definition, apiStyle string) (*Definition, error) {
```
xpofei, line 100: 为啥不直接传入 APIStyle 类型

service/rpc/builder.go
```diff
@@ -88,6 +89,9 @@ func (b *builder) SetModifier(m service.DefinitionModifier) {
 }
 
 func genRPCPath(path, version, action string) string {
+	if version == "" && action == "" {
```
xpofei, line 92: 这个判断为空返回 path 是为了 swagger 好看吗

utils/api/definitions.go
iawia002: 没注意，看见 lint 报错直接改了

service/rpc/builder.go
iawia002: 对，对于内部的一些直接用 path 形式的插件来说，log 也好看点

```
INFO  1123-11:04:57.679+08 builder.go:184 | Path: /healthz, Consumes: [*/*], Produces: [*/*]
INFO  1123-11:04:57.679+08 builder.go:184 | Path: /metrics, Consumes: [*/*], Produces: [*/*]
```

PR #402, opened 2020-11-30
fix(*): make modifier ProduceAllIfProducesIsEmpty also modify ErrorProduces

knight42:

The value of `ErrorProduces` field of `Definition` is the same as `Produces` field by default:
1. https://github.com/caicloud/nirvana/blob/b561a068104cef53a92bd475c326231132ec6158/service/rpc/builder.go#L129
2. https://github.com/caicloud/nirvana/blob/b561a068104cef53a92bd475c326231132ec6158/service/rest/builder.go#L157

Currently the modifier  `ProduceAllIfProducesIsEmpty` only modify the `Produces` field, so `ErrorProduces` would be empty If `Produces` is empty, then nirvana will throw an error when building executor:
https://github.com/caicloud/nirvana/blob/b561a068104cef53a92bd475c326231132ec6158/service/executor/executor.go#L59-L61

Fixes #

Reference to #

/cc @iawia002 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/service/modfier_test.go
+++ b/service/modfier_test.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2020 Caicloud Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


service/modfier_test.go
```diff
@@ -0,0 +1,26 @@
+package service
```
iawia002, line 1: Needs the file header here

knight42: fixed

PR #410, opened 2020-12-08
fix(utils): fix an issue when the Auto parameter is a pointer in client generation

iawia002:

Add your description

Fixes #

Reference to #

/cc @xpofei @knight42 

**Release note**:

```release-note
utils: fix an issue when the Auto parameter is a pointer in client generation.
```


knight42 (approved):
/lgtm

I have verified that this patch does address the issue.

PR #418, opened 2020-12-18
feat(swagger): support 'optional' tag for auto param

xpofei:

support 'required' tag for auto param, if defined an auto param like this:
type XXX {
	ID  int64  `source:"query,Id,optional"`
}
'required' tip will now show in swagger

Fixes #

Reference to #

/cc @iawia002 

**Release note**:

```release-note
Support 'required' tag for auto param
```

### DIFF
--- a/service/source.go
+++ b/service/source.go
@@ -18,7 +18,6 @@ package service
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"mime/multipart"
 	"reflect"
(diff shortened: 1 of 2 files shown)


service/source.go
```diff
@@ -508,13 +507,14 @@ type AutoParameterConfigKey string
 const (
 	// AutoParameterConfigKeyDefaultValue is the key of default value.
 	AutoParameterConfigKeyDefaultValue AutoParameterConfigKey = "default"
-	// AutoParameterConfigKeyAnonymous is the prefix of anonymous configs.
-	AutoParameterConfigKeyAnonymous AutoParameterConfigKey = " anonymous"
+	// AutoParameterConfigKeyDefaultValue is the key of required tag.
```
iawia002, line 510:
```suggestion
	// AutoParameterConfigKeyOptional is the key of required tag.
```

